Issue #1, opened 2019-02-05
Daily Class Schedule and Clock


maxtkc:
![crls course schedule](https://lh3.googleusercontent.com/-wbA-_r3bma0/XFjtpM4RyXI/AAAAAAAALRY/mlsMHiFNUGYNYXoqTfqpH1z8EyLA_z0wgCL0BGAYYCw/h540/2019-02-04.png)
Have the above in a better format accessible to the user.
Probably add CRLS block schedule as well, and a lunch finder/calculator (A, B, C)
 - Could be related to [crlscs.club](crlscs.club) clock
    - We could have it display current class instead of just period

maxtkc:
I am working on the clock. The general ideas is very similar to the [CRLS CS Club clock](crlscs.club), but it will read you your *class name* and instead of having numbers around the edge, we'll have a *digital countdown* of how much time is left in the current period. Also, it will use the Aspine logo as the face.

This has potential as an Android widget.

maxtkc:
Merge request created for clock #26

maxtkc:
- [x] Schedule
- [x] Clock
  - [x] Shows time between classes
  - [x] Uses schedule to identify periods
  - [x] Handle lunches (guess, and override)
  - [x] Handle weekends
  - [x] Handle black/silver days

PR #26, opened 2019-03-02
Clock dev

maxtkc:
### Gets Actual Class Names

Function added to translate `Period 1` to `<Class Name>` conditionally (if `tableData` has been loaded)

### Repositioned Class names to below the clock

While `Period n` fits nicely within the clock, `AP Computer Science Principals` is too small to read. The text has been moved out of the clock, and floats left.


maxtkc:
This should be completely ready to go. Issue #1 will be closed as well.

PR #29, opened 2019-03-17
Calendar

maxtkc:
Calendar works and is ready to be merged. There are a few additions that could be made (default calendars, etc.). Also better filtering of incorrect calendar urls.


maxtkc:
And fa fa's are still broken 😠

maxtkc:
- [x] filtering incorrect calendar id's
- [x] colors (even in settings)
- [x] ~~fa fa's work~~ not used
- [ ] ordering calendars based on number of users

PR #36, opened 2019-03-28
Fix bulleted list formatting and clarify license

psvenk:
GNU has multiple licenses, so "GNU License" is ambiguous. I changed it to "the GNU General Public License, version 3" because that is the license in the file LICENSE.md, but I would recommend changing it to "the GNU General Public License, either version 3, or (at your discretion) any later version" (or even better, "the GNU General Public License, either version 2, or (at your discretion) any later version or later" if you want to be compatible with GPLv2), so that if a GPLv4 license is released, Aspine will be compatible with projects that use GPLv4. Note that "compatible" means that it is legally possible for two projects under different licenses to share code with each other (two-way compatibility in this specific context).

I have not made this change because I have not written a single line of code in Aspine and so do not hold the copyright to the software. If you want to change the wording of the license, you should be able to push directly to my fork because I checked the "Allow edits from maintainers" box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Who needs Aspen when you have Aspine! It serves as a one stop shop for everythin
 
 ## Getting Started
 
-Using Aspine is as easy as 1, 2, 3. Just visit aspine.us and login with your Aspen credientials!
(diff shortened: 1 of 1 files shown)


psvenk:
I've closed this because there's another modification that I'd like to make before merging into `master`. I'll re-open it when it's ready.

PR #75, opened 2019-12-12
Added start and install scripts for Mac osx

notrodes:


psvenk:
We need to add an `if`-block to check if Homebrew is already installed (and only run the install script otherwise) and make sure that Homebrew does not give an error if `git`, `node`, and/or `redis` are already installed.

Also, the Homebrew install script should be checked for safety—an unsafe script can be disastrous within a `curl` if the download fails part-way, e.g. `rm -rf /usr/lib/someapp` may be truncated to `rm -rf /`. Good reading about curl to shell: [&ldquo;Piping curl to s(hell)&rdquo;](https://0x46.net/thoughts/2019/04/27/piping-curl-to-shell/) by Filip Borkiewicz; [&ldquo;Curl to shell isn&rsquo;t so bad&rdquo;](https://www.arp242.net/curl-to-sh.html) by Martin Tournoij

The install script also assumes that the user wants the Aspine git repository cloned into a subdirectory of the current directory—it would be nice if there were a prompt where the user could optionally provide a directory in which Aspine is to be cloned. Perhaps something like this, if the user's current directory is the home directory:
```
Directory in which to clone the Aspine git repository (default: ~/aspine/):
```
This should be wrapped somehow in an error-catching block so that the user is notified and prompted for another directory if the clone operation is unsuccessful, and `mkdir -pv` should be used before trying to run `git clone`.

notrodes:
All the tools we use to install do these checks for us, so the code is safe, its just like a user running it. Nothing force installs

psvenk:
In any case, we should still prompt the user for the directory in which to clone Aspine and use `mkdir -pv` to create it. It's not a good idea to keep the user (who, mind you, is a programmer) ignorant of where the repository is cloned on their system. In a similar vein, though, I don't see why an install script is needed for Mac OS X (or any Unix), where the instructions given in the `README` (with some revision) can be followed pretty easily by anyone who has some basic familiarity with the command line and willingness to debug a little bit, qualities which are required for programmers and reasonable to expect of contributors IMO (and many other projects expect a much higher level of familiarity with command-line utilities). I do understand the need in the case of Windows, because Microsoft has left a 1980s-era command interpreter as the standard and has made it very difficult to install a basic development environment under Windows.

PR #77, opened 2019-12-19
Lazy loading

### LINKED ISSUE #56: Load page while student data continues to arrive.
Add scrapers for previous quarters will increase loading time because these scraping previous quarters requires more requests than scraping the current quarter. It will therefore be more important than ever that page loads as data arrives rather than waiting for it all to arrive.

psvenk:
Closes #56. This PR moves the schedule and PDF scrapers out of the main `scrape_student` function in `scrape.js` (which is called from the `/data` endpoint) and keeps separate endpoints for them, `/schedule` and `/pdf`. `home.html` has been updated to perform separate AJAX requests for the schedule and the PDF files (reports), with items only being downloaded as needed. In the case of the schedule, the data are downloaded as soon as the grades have been downloaded or when the user clicks on the "Schedule" tab, so that the clock in the upper left-hand corner shows the current period without the user needing to click on the "Schedule" tab to load the schedule data. In the case of the reports, the data are downloaded only when the user clicks on the "Reports" tab.

From my (somewhat limited) testing, this code causes no problems. Please let me know if any of you come across any problems with the new code.


Ruborcalor:
Wow awesome work! I took care of two small problems that arose from the possibility for /pdf to return before /data. I then added a loader to indicate to users that lazy loading is going on (it could be a confusing otherwise). If you agree with these changes, I think it is ready to merge.

psvenk:
@Ruborcalor These changes look good, thank you for catching those oversights and making fixes. I did notice that the loader was gone and intended to fix it, but for some reason I forgot to do that before creating this PR. I'll just ask @kdk1616 if he has any input before merging.

PR #92, opened 2020-01-12
Update macOS install script and support multiple types of GPA (old)

### LINKED ISSUE #71: Improved GPA calculator
* [x] Percent 
* [x] Unweighted (out of 4)
* [x] Weighted (Out of 5 depending on AP, HN, or CP)
* [x] Make grade calculations work with new types of GPA (so weighted and 4.0 GPA's update when you change assignments)
* [x] Year-to-date GPA (dependent on #73)

### LINKED ISSUE #83: Revise Mac install script
- Remove bit that clones code from github.com (Instead, assume they have forked and cloned it or downloaded a zip file and then are running the script from there)
- Then download dependencies
- Separate password
- Try to make script double-clickable
- Write better instructions for README

kdk1616:
Closes #83. Closes #71. Fixed mac install and start scripts (removed redis from start script)… Made function for toggling between weighted, unweighted, and % GPA (only works for current quarter right now... Also, There is a bug where if you select another quarter you cannot return to the initial quarter and have it work properly...).


psvenk:
Looks good, but I found a few small issues, such as:
- [ ] GPA indicator should be merged with GPA type selector, not with term dropdown
- [ ] A GPA of 4.0, 3.0, etc. should be displayed with at least one place after the decimal point
- [ ] The non-percent GPA formats do not play well with "Calculated GPA"

psvenk:
From what I've seen from editing the code the past two days, it seems that it may take around a week for this to be ready to merge.

PR #93, opened 2020-01-13
Refactored package.json (old)

notrodes:
in accordance with npm docs ([here](https://docs.npmjs.com/configuring-npm/package-json.html#people-fields-author-contributors)). Moved lines, changed from author field to contributors field. Updated repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,17 @@
   "name": "aspine",
   "version": "2.0.0",
   "description": "Aspen Scraper and Alternate Frontend",
-  "main": "serve.js",
+  "homepage": "https://github.com/maxtkc/aspine#readme",
+  "bugs": {
+    "url": "https://github.com/maxtkc/aspine/issues"
+  },
+  "license": "GPL-3.0-only",
+  "contributors": ["Max Katz-Christy", "Cole Killian"],
+  "main": "src/serve.js",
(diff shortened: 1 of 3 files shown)


notrodes:
now server can be started with `npm start`

notrodes:
Dont test this now its not working

psvenk:
These changes look good; I'll test them later tonight. However, I see that you have changed double quotes to single quotes in some places. Is there a concrete reason for this? From what I have seen, most of the Aspine code uses double quotes for strings. My rationale for generally using double quotes (unless there are nested quotes) is that it is works across languages.

notrodes:
@psvenk no real reason I just decided to use single quotes from now on because it is less keystrokes and wanted stuff to match.

psvenk:
@notrodes ~~I'll amend your commit and remove the quote changes for the time being to be consistent with the rest of the code base, then.~~ *I see that the overall convention in backend code is single quotes, so I'll leave your commits as-is.* When we freeze the repo as per #96, maybe we can consider ~~changing the convention if there is a strong reason to do so~~ *imposing a uniform convention*.

notrodes:
I deleted the repository this came from.

psvenk:
Superseded by #106 because @notrodes deleted the repository from which this pull request was initiated.

PR #103, opened 2020-03-08
Fix PDF Viewer

### LINKED ISSUE #88: Only first page of PDFs shown in Reports tab
If possible, a simple fix would be just to use the controls provided by pdf.js instead of drawing our own controls.

### LINKED ISSUE #89: Fullscreen does not work in Reports tab
The PDF viewer frame retains its aspect ratio and thus only takes up about 1/3 of the screen on most monitors.

Ruborcalor:
Closes #88. Closes #89. Includes both improvements to the pdf viewer by @psvenk, and a fix for pdf scraping after the aspen update.


psvenk (approved):
Looks good to me

psvenk:
By the way, my fix for #89 only works fully in Chromium-based browsers (such as Google Chrome). The PDF viewer automatically resizes when entering but not when exiting fullscreen in Firefox, and the changes do not work in other browsers such as Pale Moon. I don't see this becoming an issue that we need to worry about, given that I would assume that the vast majority of our users visit Aspine from Google Chrome.

PR #106, opened 2020-03-17
Refactored package.json (new)

psvenk:
This is the replacement for @notrodes' original pull request #93. For some reason, he deleted his fork of the repository, so I had to create a branch on this repository and initiate a pull request from the new branch.

Original description:
> in accordance with npm docs ([here](https://docs.npmjs.com/configuring-npm/package-json.html#people-fields-author-contributors)). Moved lines, changed from author field to contributors field. Updated repository.

> now server can be started with `npm start`

PR #101, opened 2020-02-15
Update macOS scripts and support multiple types of GPA (new)

kdk1616:
This PR contains two features.
* Closes #83. Fix macOS install and start scripts, including removing Redis from the start script.
* Closes #71. Make function for toggling between weighted, unweighted, and % GPA.

PR #108, opened 2020-03-18
Clean up code

### LINKED ISSUE #96: Clean up code
Much of the codebase is poorly formatted, with inconsistent indentation and disorganization in JavaScript files. I propose that we temporarily freeze the repository at some point (stop accepting any new commits or pull requests until a set later date) in order to improve the formatting on both the frontend and the backend. The improvements may include refactoring in addition to improving formatting.

psvenk:
Closes #96. This is a pull request to house all changes made during the code freeze. The main categories of changes made are:
* Reorganize files
* Remove unneeded dependencies
* Update dependencies
* Remove unused files, functions, methods, etc.
* Reformat code


psvenk:
Checklist of things to port from @notrodes' commit https://github.com/notrodes/aspine/commit/071fd53432ddecb9d3b3ace2e555d3ec0bc174b9:

- [x] Removed test script because we do not use redis
- [X] Moved frontend dependencies to vendor directory
- [X] Moved all css pertaining to login page to login.css
- [x] Removed jsx things, unused images, unused js dependencies and login v1
- ~~Added normalize.css and main.css from html5 boilerplate~~
- [x] changed htmls classes that were used once to ids
- ~~Renamed manifest to be in line with html5 boilerplate~~

psvenk:
Some additional things that will be nice to do before ending the code freeze (we can add more things to this list as needed):
* [x] Use node modules to create a virtual `vendor` directory. This way, we don't need to worry about keeping dependencies updated and we don't need to distribute dependencies with the project. [This excellent answer on Stack Overflow explains an easy way to do this with Express](https://stackoverflow.com/a/27464258).
Note: we cannot symlink subdirectories of `node_modules` into `public/vendor` because git treats symlinks as text files on Windows.

public/login.html
```diff
@@ -105,7 +105,6 @@
         scale: 1.1
       })
     </script>
-    <script src="js/main.js"></script>
```
Ruborcalor: Why do we no longer need input validation? For context the commit 
a0059b2 mentioned main.js being no longer necessary because input validation was no longer necessary.

serve.js
```diff
@@ -73,6 +73,55 @@ if(!(args.hasOwnProperty('insecure') || args._.includes("insecure"))) {
     }
 }
 
+// Expose frontend dependencies from node-modules
+// https://stackoverflow.com/a/27464258
+new Map([
```
Ruborcalor, line 78: sweet this is very cool

Ruborcalor (approved):
Looks brand new!

public/login.html
psvenk: @Ruborcalor: @notrodes and I thought that input validation is not being used anywhere because the input doesn't seem to be validated at all (`main.js` seems to check for an email format, but non-email usernames are accepted). I haven't been able to trigger the input validation on the production server (https://aspine.cpsd.us). I'm happy that you brought this up because I'm a bit confused about the input validation. If you could provide further insight, that would be great.

serve.js
psvenk: Thank you

psvenk:
By the way, the main reason why this is still a "draft" pull request and is not yet marked as "ready for review" is that I thought that it might be a good idea for this to be evaluated (to go on production) separately from the changes in the current `staging` (#101, #103, and #106) because those changes deal with behavior while these changes deal with formatting, and because the combined diff for these changes is huge and would completely dwarf those changes, making it difficult for the combined "batch" of changes to be audited and approved for production.

public/login.html
Ruborcalor: I agree that main.js doesn't appear to do any useful validation. Removing it seems to be the right thing to do, but maybe we can add an issue for input validation on the login page that would be resolved in a future pull request.

psvenk: The main issue that I see is that the district could change the format of usernames on Aspen at any time and Aspine could simply not work for those people whose usernames are in the new format. While not changing the number of digits, the district has switched the first digit of student IDs and has given some teachers non-standard usernames. It might also be useful to allow any type of username in case we decide to expand later to support teacher accounts (I do know at least one teacher who may be interested).

Ruborcalor: Ok sure that makes sense

psvenk:
This seems ready to merge now (based on @Ruborcalor's review and the sentiments of @notrodes and me), but I'll keep it as a "draft" for now due to the issues I mentioned in my previous comment.

PR #118, opened 2020-04-15
Add version number and add message for no-JS visitors

psvenk:
Add an endpoint `/version` that gets the current version of Aspine from `git describe`, display this version on the home page, and add a `<noscript>` tag with a message telling visitors with JavaScript disabled that Aspine requires JavaScript to be enabled.

Using `git describe` to get the version number instead of just reading from `package.json` has the advantage that changes made after a version is tagged are incorporated into the version number. For example: when on version 2.2.1, `git describe` outputs `v2.2.1`, but when on the branch `version` from which this PR was initiated, `git describe` outputs `v2.2.1-4-gc9803ee` because the branch is 4 commits ahead of `v2.2.1` and the latest commit is `gc9803ee`.

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -1,0 +1,51 @@
+# Maintenance
+
+This file is meant for noting information relating to project maintenance
+so that maintainers have a written record of key procedures.
+
+## Tagging a release
+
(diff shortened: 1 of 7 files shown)


Ruborcalor (approved):
Looks good

Issue #74, opened 2019-11-23
Client-side import/export feature


psvenk:
Allow users to download an archive of their grades for a year as a JSON file, choosing what to include (drawing from the global variable `tableData`). This is meant as a (much more user-friendly) improvement to the changes made in #62, which no longer work as intended due to changes made to the backend for lazy loading (#77).

Additionally, include a feature in which users can upload a JSON export and have Aspine (on the client side) intelligently merge the contents of the export with the current contents of `tableData`, overwriting only what is needed.

It would also be nice to have an option at the login page to enter Aspine with a blank `tableData` and upload whatever data one wants to use instead of scraping Aspen. This option would be useful for graduated students to view their grades by being able to use Aspine to view their exports without having an account on the Aspen server.

We would need to include a version field (and possibly a commit hash, too) in the generated JSON file which matches the current version in `package.json` (obviously, this depends on having a real versioning scheme). If a user uploads a JSON file from an earlier version, we could display a warning, and if a user uploads a JSON file from an earlier major version (not backwards-compatible), we could direct the user to our Releases page on GitHub to download and host an older version of Aspine to view their export (or, once we reach that stage, we could make a purely client-side version of Aspine which only supports JSON upload due to lacking a backend; this can be automated at each release using a specialized build process that includes all necessary dependencies and is not reliant upon any endpoints).

psvenk:
This is mostly implemented (in the `json` branch). A few things to do before opening a pull request for review:
- Further testing with incomplete exports; i.e., an export might not include all information.
  + [x] It may be beneficial to require that certain data are included in every export (e.g. grades overview for ease of calculating GPA) so that we do not have to deal with edge cases throughout the codebase. - *Fixed in 3c02e715f5.*
- Further testing with merging an export with scraped data- for example, if quarter 1 and quarter 3 grades were scraped and quarter 2 grades are included in the export, they should all be displayed in the term dropdown.
[...]

PR #119, opened 2020-04-21
Client-side JSON import/export

psvenk:
Implements #74 at long last. Please see that issue and the commit messages for more details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 db/*.rdb
 usage_log.txt
 .DS_Store
+dist-lite/
--- a/build-lite.sh
+++ b/build-lite.sh
@@ -62,6 +62,6 @@ dist-lite/js/clock.js
 # Include contents of schedule.json
 # https://unix.stackexchange.com/a/32912
 
-version="$(git describe)"
+version="$(git describe | sed 's/^v\?\(.*\)/\1/')"
 sed -i 's/\/\/#include $version/"'$version'"/g' dist-lite/js/*
-# Hard-code version number
+# Hard-code version number (use sed to trim 'v')
--- a/build-lite.sh
+++ b/build-lite.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+# Build script for a "lite" (purely client-side) version of Aspine
+
+# Note: This script has only been tested with Bash in a GNU environment
+# (on Linux); your mileage may vary on BSD, macOS, Cygwin/MSYS (Windows), etc.
+
+rm -r dist-lite/
+mkdir -p dist-lite/
+
+cp -r public/* dist-lite/
+
+mappings="$(
+grep -A100 'new Map' serve.js \
+| grep -A3 '^    \[' \
+| grep -Ev '\[|\]' \
+| sed 's/        '\''//' \
+| sed 's/'\'',*//'
+)"
+# Find the line in serve.js containing 'new Map' and get the next 100 lines
+# Limit the search to lines that contain '    [' and the next three lines
+# Remove lines with brackets
+# Remove leading whitespace and single quote character
+# Remove trailing single quote character and comma (if any)
+
+echo $mappings | \
+while read -d ' ' endpoint
+do
+    read -d ' ' path
+
+    mkdir -p "$(dirname dist-lite$endpoint)"
+    # Create the parent directory
+    cp ".$path" "dist-lite$endpoint"
+    # Copy the node module to dist-lite
+done
+# https://stackoverflow.com/a/21256704
+
+cp -r ./node_modules/@fortawesome/fontawesome-free/webfonts/ dist-lite/fonts/fontawesome/webfonts
+# Copy Font Awesome files
+
+rm dist-lite/login.html
+# Remove login page
+
+sed '/Logout/d' -i dist-lite/home.html
+# Remove "Logout" button
+
+sed -e '1h;2,$H;$!d;g' -e 's/\
+\$\.ajax({\
+\ \ \ \ url:\ "\/data.*then(responseCallback);/\
+responseCallback({\
+    classes: [],\
+    recent: {\
+        recentActivityArray: [],\
+        recentAttendanceArray: []\
+    },\
+    overview: [],\
+    username: "",\
+    quarter: "0"\
+});/' \
+-i dist-lite/js/home.js
+
+sed -e '1h;2,$H;$!d;g' -e 's/\
+\$\.ajax({\
+\ \ \ \ url:\ "schedule.json.*then(schedulesCallback);/\
+schedulesCallback(\
+);/' \
+-i dist-lite/js/clock.js
+
+sed -e '/schedulesCallback(/r dist-lite/schedule.json' \
+-i dist-lite/js/clock.js
+
+# Replace AJAX calls with initialization using blank data
+# https://unix.stackexchange.com/a/235016
+# https://unix.stackexchange.com/a/32912
+
+version="$(grep version package.json | sed 's/.*: "//' | sed 's/",//')"
+# Get version number from package.json
+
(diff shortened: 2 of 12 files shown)


psvenk:
Some things to do before merging (from my conversation with @kdk1616):
- [x] On the login page, say "Use Aspen credentials or leave blank **to import data**"
- [x] Add a mouseover to each disabled term in the export dialog stating the reason why the term is disabled (term is not available on Aspen, or term is not included in the imported data)
- [x] Grey out inaccessible quarters in the quarter dropdown

psvenk:
I just now found a bug: if more than one object is in tableData (e.g.
scraped data and imported data), then the quarter dropdown does not
adjust on switching currentTableData. A fix is underway.

psvenk:
Also, further testing is needed regarding how switching `currentTableData` affects calculated grades (see `termsReset`) and different formats of GPA.

Ruborcalor:
Great stuff! Some ideas.

1. When logging in and then importing data, the term selector doesn't fully reset.  In the screenshot below I imported only two terms, but all four still appear available in the term selector. What do you think about indicating which ones were included in the import?
<details>
 <summary>Term Selector</summary>

</details>

2. The default behavior seems to be that when navigating to home.html, you are prompted for an import file, however I imagine that most users would prefer to be redirected back to the login.html page if their session timed out.

psvenk:
@Ruborcalor Thanks for the review.

On 4/21/20 6:09 PM, Cole Killian wrote:

> Great stuff! Some ideas.
> 
>  1. When logging in and then importing data, the term selector doesn't
>     fully reset. In the screenshot below I imported only two terms, but
>     all four still appear available in the term selector. What do you
>     think about indicating which ones were included in the import?

I think this is caused by two of the issues that I mentioned above:

> - Grey out inaccessible quarters in the quarter dropdown
>
> - If more than one object is in tableData (e.g. scraped data and
>   imported data), then the quarter dropdown does not adjust on switching
>   currentTableData.

Sorry, I may not have worded them very well. Hopefully the issue that
you noticed will be resolved when my fix is done.

>  2. The default behavior seems to be that when navigating to home.html,
>     you are prompted for an import file, however I imagine that most
>     users would prefer to be redirected back to the login.html page if
>     their session timed out.

I imagined that most users would navigate to the / endpoint (which now
redirects since faed11dedf to the correct page depending on whether or
not there is an active session) without accounting for the fact that
people may have bookmarked /home.html.

This can probably be fixed by redirecting home.html to login.html
server-side when there is no active session, and creating a dummy
session when a user clicks on "Login" leaving the username and password
fields blank.

psvenk:
> - On the login page, say "Use Aspen credentials or leave blank **to import data**"
> - Add a mouseover to each disabled term in the export dialog stating the reason why the term is disabled (term is not available on Aspen, or term is not included in the imported data)

Fixed in 7ec1869a3f.

> - Grey out inaccessible quarters in the quarter dropdown

Fixed in 89a9c67106.

> - If more than one object is in tableData (e.g. scraped data and imported data), then the quarter dropdown does not adjust on switching currentTableData.

Fixed in 4e87e1e114.

Remaining things to do:

- > The default behavior seems to be that when navigating to home.html, you are prompted for an import file, however I imagine that most users would prefer to be redirected back to the login.html page if their session timed out.
- Instead of using the HTML `title` element for tooltips, use CSS and/or JavaScript to make a more accessible alternative that works on mobile devices and works with assistive technologies. See [MDN][1] and [Inclusive Components][2] for more info.
  + This is a good opportunity to mention that the accessibility of Aspine is not optimal and that we should add semantic attributes (among other improvements) at some point.
- Fix `build-lite.sh` to work properly after 05435d62b8.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title
[2]: https://inclusive-components.design/tooltips-toggletips/

psvenk:
Approval has been secured from @notrodes

notrodes (approved):
Yeah good work dudeman.

psvenk:
Another thing to fix: The version detection for imports currently only checks the major version. We need to warn the user if the minor version of the import is newer than the minor version of Aspine, letting them know that they should use a newer version of Aspine to view the file (presumably Aspine lite).

psvenk:
I would like to note that the solution employed in 89a9c67106 is hacky in that it uses `setTimeout` to defer loading the term dropdown, and we should aim for a more sustainable solution in the future.

psvenk:
Note to self re hackiness of solution: this can be fixed by rewriting `isAccessible` to take an optional argument which specifies which terms are included in an import (in the case of imported data) and reads that argument if it is present.

We can probably get away with not passing in anything like that for scraped data as long as `currentTableData.overview` is populated. IIRC that is the only field that is checked for scraped data.

psvenk:
> I would like to note that the solution employed in [89a9c67](https://github.com/Aspine/aspine/commit/89a9c6710640b356518395f3a025ca10239aac31) is hacky in that it uses `setTimeout` to defer loading the term dropdown, and we should aim for a more sustainable solution in the future.

Fixed in d939809c43.

psvenk:
> The default behavior seems to be that when navigating to home.html, you are prompted for an import file, however I imagine that most users would prefer to be redirected back to the login.html page if their session timed out.

Fixed in 3393b11583.

psvenk:
> Fix `build-lite.sh` to work properly after [05435d6](https://github.com/Aspine/aspine/commit/05435d62b8b07c67889f7398b0f118764f02d9b5).

Fixed in d61f403a24.

Remaining issues:

> Instead of using the HTML `title` element for tooltips, use CSS and/or JavaScript to make a more accessible alternative that works on mobile devices and works with assistive technologies. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) and [Inclusive Components](https://inclusive-components.design/tooltips-toggletips/) for more info.

> The version detection for imports currently only checks the major version. We need to warn the user if the minor version of the import is newer than the minor version of Aspine, letting them know that they should use a newer version of Aspine to view the file (presumably Aspine lite).

psvenk:
> Instead of using the HTML `title` element for tooltips, use CSS and/or JavaScript to make a more accessible alternative that works on mobile devices and works with assistive technologies. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) and [Inclusive Components](https://inclusive-components.design/tooltips-toggletips/) for more info.

Fixed in db0604dfa6.

psvenk:
> > Instead of using the HTML `title` element for tooltips, use CSS and/or JavaScript to make a more accessible alternative that works on mobile devices and works with assistive technologies. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) and [Inclusive Components](https://inclusive-components.design/tooltips-toggletips/) for more info.
> 
> Fixed in [db0604d](https://github.com/Aspine/aspine/commit/db0604dfa667c36dc7cfebf04f368bffd522ddcd).

Sorry, wrong issue

psvenk:
> The version detection for imports currently only checks the major version. We need to warn the user if the minor version of the import is newer than the minor version of Aspine, letting them know that they should use a newer version of Aspine to view the file (presumably Aspine lite).

Fixed in db0604dfa6.

psvenk:
> Instead of using the HTML `title` element for tooltips, use CSS and/or JavaScript to make a more accessible alternative that works on mobile devices and works with assistive technologies. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) and [Inclusive Components](https://inclusive-components.design/tooltips-toggletips/) for more info.

Fixed in 847ba33471.

All of the blockers that have been identified so far have been fixed.

@Ruborcalor @kdk1616 @notrodes Please take a look. Is this ready to merge?

Ruborcalor (approved):
Looks good to merge

kdk1616 (approved):

PR #121, opened 2020-05-04
Change build-lite.sh to POSIX sh syntax

psvenk:
This should solve incompatibilities with non-GNU systems.

@kdk1616, would you check if this works on macOS?

Windows (without Cygwin/MSYS/WSL) is not supported.

### DIFF
--- a/build-lite.sh
+++ b/build-lite.sh
@@ -1,38 +1,36 @@
-#!/bin/bash
+#!/bin/sh
 # Build script for a "lite" (purely client-side) version of Aspine
 
-# Note: This script has only been tested with Bash in a GNU environment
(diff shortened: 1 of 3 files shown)


psvenk:
The script works flawlessly on a clean install of OpenBSD, but I think macOS uses outdated versions of BSD utilities, so @kdk1616's testing will still be valuable.

psvenk:
I wanted to test with the old version of Bash included in macOS, but apparently it is from 2006 and predates Node.js. It seems that I will need to scp the node_modules directory...

psvenk:
Works on Ubuntu 7.04.

Ruborcalor (approved):

PR #125, opened 2020-06-07
Re-implement stats plot in d3 + other optimizations

### LINKED ISSUE #122: Use a lighter library instead of Plotly.js
Currently, we use Plotly.js in its default distribution (the package `plotly.js-dist` on NPM), which ships a 7.2M (1.7M gzipped) bundle! The partial bundle in the package `plotly.js-cartesian-dist-min` is 1.0M (343K gzipped), which is much better but still very heavy. I also tried creating a custom partial bundle (minified, of course) with only the box plot functionality (because that is the only functionality that we use), but this was still 690K (235K gzipped) and I haven't been able to include it in Aspine without also pulling in the `devDependencies` of Webpack and the full Plotly.js source code.

Considering that we only use Plotly.js for fairly lightweight box plots, maybe we should look into a more lightweight and simple library. As a stopgap solution, the `lean` branch uses the comparably light `plotly.js-caresian-dist-min` bundle.

Alternatives to Plotly.js:
- [D3][1]: 243K (80K gzipped) - this is actually the library that powers Plotly.
  - [d3-boxplot][2]: 4.2K (1.6K gzipped)
- [Chart.js][4]: 169K (52K gzipped)
  - [chartjs-chart-box-and-violin-plot][5]: 21K (6.2K gzipped)
- [Flot][3]: 104K (33K gzipped)
  - Box plot support unknown
- [jqPlot][6]
  - Box plot support unknown
- [Chartist.js][7]
  - Box plot support unknown

[1]: https://github.com/d3/d3
[2]: https://github.com/akngs/d3-boxplot
[3]: https://github.com/flot/flot
[4]: https://github.com/chartjs/Chart.js
[5]: https://github.com/datavisyn/chartjs-chart-box-and-violin-plot
[6]: http://www.jqplot.com/
[7]: https://github.com/gionkunz/chartist-js

psvenk:
Closes #122. Plotly.js is a very heavy library, so I have re-implemented the stats plot in d3, a much lighter and more modular library for graphs.

| | Total assets | (gzipped) |
| - | - | - |
| Before | 6279K | 1734K |
| After | 1376K | 482K |

While 0.99M (335K gzipped) of the "Before" is Plotly alone, the sum of the d3 components in "After" is only 71.3K (24.7K gzipped)&nbsp;&mdash;&nbsp;the bulk of the 1376K (482K) figure comes from PDF.js, Tabulator, and the Poppins font.

Screenshots:
<details>
<summary>
Before
</summary>

image

</details>

<details>
<summary>
After
</summary>

image

</details>


psvenk:
I just now noticed that the textual display of the statistics is under the "Graph" header. Maybe it would be better just to do away with the "Graph" and "Info" headers entirely.

Ruborcalor:
Great work thank you for taking care of this.

I agree that it doesn't make sense to put text under a header titled "Graph". What do you think of moving the info into a table like this? (I didn't push these changes yet)

image

psvenk:
On 6/11/20 9:44 PM, Cole Killian wrote:
> I agree that it doesn't make sense to put text under a header titled "Graph". What do you think of moving the info into a table like this? (I didn't push these changes yet)
> 
> <details>
> <summary>
> Screenshot
> </summary>
>
> image
>
> </details>

@Ruborcalor Looks great! One small thing: I think there is not enough contrast between the columns. Maybe we could bold the left column and/or use different background colors for the left and the right (using borders to separate rows)?

Ruborcalor:
Cool I pushed a version with a bold left column. Feel free to play around with the layout and style.

Ruborcalor (approved):

PR #127, opened 2020-06-17
Test corrections

kdk1616:
I implemented the test corrections feature to the assignmentsTable. It allows users to apply test corrections to assignments without having to do the math themselves. 

Bugs to fix before implementing:

1) I couldn't get the icon in the top bar of the table to look right no matter what I did (HELP)
2) The only way I got the modal to work for this was to "hijack" the stats modal. Presumably, this is not ideal...

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -353,7 +353,8 @@ i.fa.fa-plus.grades {
   cursor: pointer;
 }
 
-i.fa.fa-info-circle {
+i.fa.fa-info-circle,
+i.fa.fa-toolbox {
   color: #00551C;
   padding-left: 31%;
   padding-top: 25%;
--- a/public/home.html
+++ b/public/home.html
@@ -151,6 +151,19 @@
     </div>
   </div>
 
+  <div id="corrections_modal" class="modal">
+    <!-- Modal content -->
+    <div id="corrections_modal_content" class="modal-content">
+      <span class="close" onclick="hideModal('corrections')">&times;</span>
+      <h3>Test Corrections</h3>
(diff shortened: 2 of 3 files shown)


psvenk:
@kdk1616 Thanks for the feature and PR. This looks almost ready to merge.

Changes that I made:
- Rebase on master
- Fix up formatting
- Change `tableData` to `currentTableData` pursuant to the changes in version 2.3.0
- Make modal code less hacky by giving test corrections its own modal instead of hijacking the stats modal

Changes that I plan on making:
- Make form code (currently using a HTML form and an iframe to bind the Enter key to the button) less hacky
- Remove the (misaligned) toolbox icon at the header (it is not a button anyway, and anyway this is to be consistent with the stats button)
- Use original score instead of calculated score

Also, as you said, we should look into the better way to put buttons in a table with Tabulator. What we are doing (overall in the UI) is probably an unsupported hack because some parts look misaligned when I tried to update Tabulator to another minor version (which is not supposed to have breaking changes).

I plan on squashing all changes into one commit because this is a relatively small change and the diff would be pretty readable.

psvenk:
I have made all of the above changes (except for removing the toolbox icon; apparently the stats button uses such an icon in the header as well, so I just added some CSS to make it look better). @kdk1616, would you please take a look? Pressing Enter to select the button to apply test corrections works on my end without need for any event handlers.

psvenk (approved):

psvenk:
I just now added autofocus to the text box and bound the Enter key in the text box to applying corrections. @kdk1616 is this ready to merge?

PR #128, opened 2020-06-27
Adding documentation

notrodes:
this is not ready for a merge, just let me know if this is a good idea!

### DIFF
--- a/public/js/buttonFunctions.js
+++ b/public/js/buttonFunctions.js
@@ -368,7 +368,9 @@ let exportTableData = async function(prefs) {
 };
 
 /**
- * @param {TableDataObject} obj
+ * Imports a JSON file and adds the data as a TableDataObject to the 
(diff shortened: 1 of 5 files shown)


psvenk:
This seems like a good idea. I would put the `typedef`s in a separate `types.ts` file, specified using regular TypeScript syntax (`export type TableDataObject = { ... }`), and have one import block like [this][1]:
```javascript
/**
 * @typedef {import("./types").TableDataObject} TableDataObject
 */
```

[1]: https://github.com/flippeducation/flippeducation/blob/dfbd86ad2f3ad534918638a3b007e212e96dc9ed/src/index.js#L11

psvenk:
TODO (when reviewing this for merge): remove merge from origin to feature branch so that history is cleaner

psvenk (approved):

PR #149, opened 2020-10-24
Removed synthetic statistics

### LINKED ISSUE #148: Statistics for synthetic assignments
Added assignments currently have an info button and try to display statistics but aren't real and don't have any statistics tied to them. They just display data from the last assignment whose statistics have been looked at. 

To fix this we should probably tag synthetic assignments somehow and just not show an info button on them. I'm currently thinking about adding the tabulator row index of the added assignments to a list to tag them.

andOrlando:
Closes #148. Here's the pull request. I forgot to do this last time. The only thing I changed was removing the info button and it's click event from synthetic assignments.

### DIFF
--- a/public/js/buttonFunctions.js
+++ b/public/js/buttonFunctions.js
@@ -11,6 +11,7 @@ let newAssignment = function() {
       "max_score": 10,
       "percentage": 100,
       "color": "green",
+      "synthetic": "true",
     });
     
     updateGradePage();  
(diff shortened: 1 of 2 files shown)


psvenk:
@andOrlando Thank you for your contribution. I noticed that `total_synthetic_assignments` is a global variable which applies to all classes in all quarters, meaning that if the user creates a synthetic assignment in one class and then selects another class, the statistics for the top assignment in the other class become inaccessible. I just now pushed a commit that adds an attribute `"synthetic": true` to synthetic assignments and checks against that attribute instead of using a global counter.

psvenk (approved):

PR #150, opened 2020-10-25
Distinguish levels within letter grades by using gradients

### LINKED ISSUE #50: Different Colors for Different Percentages Within a Letter Grade
Differentiate, for example, between an 80% and an 88%.

Preferably via a gradient.

andOrlando:
would close #50 
doesn't use the exact values from the color scheme but it definently fits. Also all of the colors are just constants and can be changed at any time, possibly could be put in settings at a later date.

There were two ways I did this:
First one is a graident between two distinct colors which you can pick by editing the constants
Second one is a gradient from one color to black, but the color varies based on what grade range you're in using colors similar to the old verison.

Pick whichever you think fits best, I'm personally in favor of the latter and its closer to what we currently have.

### DIFF
--- a/public/js/extraFunctions.js
+++ b/public/js/extraFunctions.js
@@ -221,7 +221,7 @@ const black_rgb = [0, 0, 0]
 function calculateColorAddition(max, min, grade, color_rgb) {
   let result = "#"
   const percentage = (max-grade)/(max-min)/2
-  for (i=0; i<3; i++) {
+  for (const channel of color_rgb) {
     const color = Math.round(color_rgb[i]-percentage*color_rgb[i]).toString(16)
     result += color.length == 1 ? "0" + color : color
   }
--- a/public/js/extraFunctions.js
+++ b/public/js/extraFunctions.js
@@ -210,22 +210,22 @@ function getLetterGrade(gradeToBeLettered) {
 
 
 //consts for getColor
-const green_rgb = [29, 219, 93]
-const blue_rgb = [102, 102, 255]
-const yellow_rgb = [255, 231, 0]
-const orange_rgb = [255, 165, 0]
-const red_rgb = [255, 0, 0]
-const black_rgb = [0, 0, 0]
+const green_rgb = [29, 219, 93];
+const blue_rgb = [102, 102, 255];
+const yellow_rgb = [255, 231, 0];
+const orange_rgb = [255, 165, 0];
+const red_rgb = [255, 0, 0];
+const black_rgb = [0, 0, 0];
 
 //calculates how much color should be added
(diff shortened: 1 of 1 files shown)


notrodes:
is this needed?

public/js/extraFunctions.js
```diff
@@ -208,24 +208,46 @@ function getLetterGrade(gradeToBeLettered) {
...
+const yellow_rgb = [255, 231, 0]
+const orange_rgb = [255, 165, 0]
+const red_rgb = [255, 0, 0]
+const black_rgb = [0, 0, 0]
+
+//calculates how much color should be added
+function calculateColorAddition(max, min, grade, color_rgb) {
+  let result = "#"
+  const percentage = (max-grade)/(max-min)/2
+  for (i=0; i<3; i++) {
```
notrodes:
```suggestion
  for (const channel of color_rgb) {
```

public/js/extraFunctions.js
```diff
@@ -208,24 +208,46 @@ function getLetterGrade(gradeToBeLettered) {
...
+    const color = Math.round(color_rgb[i]-percentage*color_rgb[i]).toString(16)
```
notrodes:
```suggestion
    const color = Math.round(channel-percentage*channel).toString(16);
```

public/js/extraFunctions.js
```diff
@@ -208,24 +208,46 @@ function getLetterGrade(gradeToBeLettered) {
   }
 }
 
+
```
notrodes, line 211: ### semi-colons

notrodes (approved):
Ok

public/js/extraFunctions.js
```diff
@@ -208,24 +208,46 @@ function getLetterGrade(gradeToBeLettered) {
   }
 }
```
notrodes, line 209: ooops I made mistake, fix these first

notrodes (changes requested):

notrodes (approved):

psvenk:
I'll take a look at this tomorrow and merge if there are no objections or questions.

psvenk (changes requested):
- Grades close to 100% are given a very bright green color which is not very pretty.
- When creating an assignment, the initial color is the default green and not the color that corresponds to a grade of 100% (the default upon creating an assignment).

psvenk:
One idea that was brought up yesterday was that we abandon this pull request (sorry @andOrlando) and instead use the CSS color variables for the colors that are used for grades, because this would make it easier to adapt grade coloring to dark mode (see #195).

andOrlando:
NOOOOoOoOOOOOoOooooooooooo

PR #155, opened 2020-10-28
switch to 'aspine listening on port' instead of 'example app'

MohanMaker:
Changed serve.js file to print 'Aspine listening on port' instead of 'Example app.'


psvenk (approved):
Wow, that was quick. Thanks @MohanMaker!

MohanMaker:
Thanks for merging @psvenk :)

PR #157, opened 2020-10-30
Add an info tab + informative tooltips

### LINKED ISSUE #76: Add an instructions page
Add an "Instructions" page with advice for how to best take advantage of Aspine's features. The features are sometimes unclear to new users. This page could potentially even take the form of a video.

Potential instructions:

* On the recent activity page: Using the switch to view "Assignments" or "Attendance"
* On the schedule page: using the switch to view "black" days or "silver" days
* On the grades page, taking advantage of the "info" icon for viewing statistical data and teacher feedback.
* How to use the grade prediction feature (eventually including the percent test corrections feature from #72)

### LINKED ISSUE #178: Show descriptive text when hovering over icons to improve user experience
Discussed at our meeting last week. Allow a small bit of descriptive text to be shown when you hover over an icon. For example, many people don't know what the hammer icon for editing assignments means, so it could say "test corrections" or "edit assignment" when you hover over it.

andOrlando:
Closes #76. Closes #178. Please do fix any inconsistancies/things that go against convention and tell me about it becuase I haven't really done too much html. We can probably improve on this as people have ideas and with this tab we really do need a dropdown or something before we actually merge this as to not have way too many tabs.

Also I don't know how to collaborate on a branch so I'm just going to pull I guess

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -131,8 +131,14 @@ h4#title {
 .info-header {
   font-family: Poppins-Bold, Arial, Helvetica, sans-serif;
   font-weight: bold;
-  font-size: 18px;
-  color: #00551D
+  font-size: 1.25em;
+  color: #00551D;
+}
+
+.info-text {
+  font-size: 1em;
+  color: #000000;
+  margin-bottom: 1.5em;
 }
 
 .tabulator-col {
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -145,6 +145,10 @@ h4#title {
   height: 35px;
 }
 
+.tabulator-col.icon-col {
+  text-align: center !important;
+}
+
 button {
   font-family: Arial, Helvetica, sans-serif;
   font-weight: bold;
@@ -388,8 +392,8 @@ div.tabcontent, div.tab {
 
 i.fa.fa-plus.grades {
   color: var(--green2);
-  padding-left: 31%;
-  padding-top: 25%;
+  /* padding-left: 31%; */
+  padding-top: 8.5px;
   font-size: large;
   cursor: pointer;
 }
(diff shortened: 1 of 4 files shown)


psvenk:
I am reopening this because it is good to have this in a separate pull request than #164.

kdk1616:
Add hover over instructions to features on home screen. (i.e. on test-corrections or download buttons)

andOrlando:
- add ICTS involvement?
- add the github and privacy policy link
- add GPA
- add import/export
- fix grammar and spelling, change wording (service → tool, thusly → thus)

psvenk:
@andOrlando Thanks for the changes. Now that I see the custom tooltip, I actually think that it looks nicer than simply using the HTML `title` attribute (though I think `font-family: sans-serif` for tooltip text would look better than using Poppins-Regular).

Having tooltips reminds me that the icons in the header of the grades table are confusing because the "Add assignment" icon is a button while the other two (test corrections and statistics) are not, and because the per-assignment icons directly below the plus sign are for deleting assignments. We could make it so that each individual button (for each assignment) has a tooltip and do away with the decorative header icons (so that the plus sign is the only icon in the header). Your thoughts?

I also noticed that there is quite a bit of boilerplate for each tooltip. There is a pure-CSS way to read from an attribute of an element (instead of having multiple elements), so we should be able to achieve the same styling with less HTML code (see <https://stackoverflow.com/q/2011142> and <https://jsfiddle.net/z42r2vv0/2/>) by reading from the `title` attribute or something else like `data-title` (e.g., if there is a double-tooltip effect after hovering for long enough).

andOrlando:
@psvenk Each individual button with a tooltip sounds cumbersome, whereas the green tint of the plus button makes it pretty clear what it does, and in my opinion it'd probably be fine just to let people figure out that it refres to the icon buttons. 

I will make the attribute changes asap, didn't know those things existed.

Edit: Oops I thought the + was more green, I might change that in a separate commit and people can say what they think wednesday

psvenk:
@andOrlando Instead of having an instantly appearing tooltip for each individual button, we could use a CSS transition to make the tooltips appear only after the user hovers over a button for more than, e.g., 1 second. Alternatively, changing the color of the statistics and test corrections header icons from green to a more neutral color like gray, combined with removing the tooltip from the plus button, could make things more clear and remove the need for tooltips on individual buttons.

psvenk (commented):
Looks good to me, but let's hold off on merging for now until some other people take a look and comment on the length of the delay for the tooltips.

psvenk:
Superseded by #204 and #205.

PR #162, opened 2020-11-02
"Add assignment" button makes it add to current category

andOrlando:
adds to buttonFuntions.js and home.js, changing the newAssignment function so it adds to the current category filter (and if none is selected just the first category filter), and tracks what the filter is currently. I did this based off a variable and not the filter because it's easier. Just a small bug kind of.


andOrlando:
also forgot to mention this, but selecting the same category as the one currently selected clears the filter

psvenk (approved):
Looks good to me!

PR #164, opened 2020-11-07
Sidenav based on screen resize

### LINKED ISSUE #152: Collapse navigation bar into hamburger menu on small screens
Continuation of <https://github.com/Aspine/aspine/issues/144#issuecomment-715927138>. The tabs on the navigation bar take quite a bit of space on mobile displays, so we should collapse them into a hamburger menu depending on screen size (whether through CSS `@media` queries or through JavaScript).

<details>
<summary>Screenshot of the navigation header taking up 4 rows on a 640x1136 mobile screen</summary>

IMG_5226

</details>

andOrlando:
Closes #152 
Adds a hamburger with different contents based on screensize. Also comes with a couple css changes for dropdown-related things. For more specific details, see commits

~~Importantly, this doesn't work dynamically so should we add tabs or change tab sizes, we're going to reconfigure the exact screensizes in both the css and the javascript~~ fixed that

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -76,10 +76,10 @@
   <!-- SIDENAV STUFF -->
   <div id="sidenav" class="sidenav">
     <!-- switch-exempt makes the switch out of navbar pass over this -->
-    <button onclick="closeSideNav()" class="tablinks switch-exempt">
-      <i class="fa fa-times" style="position: absolute; top: 20px; left: 20px;"></i>
-      Close
+    <button onclick="closeSideNav()" class="tablinks switch-exempt" style="margin: 0px;width: 270px; height: 44px; color: #eee">
+      <i class="fa fa-times fa-2x" style="position: absolute; top: 18px; left: 18px; font-size: 1.5em; color: #000""></i>
     </button>
+    <hr/>
     <div class="GPA tableData_custom-select custom-select tablinks-right", id="navbar_tableData_div" style="width: 270px;">
       <select id="tableData_select">
         <option value="0">Current Year</option>
@@ -92,16 +92,16 @@
   </div>
 
   <div class="tab">
-    <button class="tablinks" onclick="openTab(event, 'grades')" id="default_open">Grades</button>
-    <button class="tablinks" onclick="openTab(event, 'schedule')">Schedule</button>
-    <button class="tablinks" onclick="openTab(event, 'recent')">Attendance</button>
+    <button class="tablinks hide switch-exempt" onclick="openSideNav()" id="hamburger_button"><i class="fa fa-bars"></i></button>
+    <button class="tablinks" onclick="openTab(event, 'grades'); closeSideNav()" id="default_open">Grades</button>
+    <button class="tablinks" onclick="openTab(event, 'schedule'); closeSideNav()">Schedule</button>
+    <button class="tablinks" onclick="openTab(event, 'recent'); closeSideNav()">Attendance</button>
     <!--<button class="tablinks" onclick="openTab(event, 'clock')">Clock</button>-->
     <!-- TODO: post-covid, setting lunch should be migrated to settings -->
     <!--<button class="tablinks" onclick="openTab(event, 'calendar-tab')">Calendar</button>-->
-    <button class="tablinks" onclick="openTab(event, 'reports')" id="reports_open">Reports</button>
-    <button class="tablinks" onclick="openTab(event, 'information')" id="information_open">Info</button>
+    <button class="tablinks" onclick="openTab(event, 'reports'); closeSideNav()" id="reports_open">Reports</button>
+    <button class="tablinks" onclick="openTab(event, 'information'); closeSideNav()" id="information_open">Info</button>
 <!--#ifndef lite-->
-    <button class="tablinks tablinks-right hide" onclick="openSideNav()" id="hamburger_button"><i class="fa fa-bars"></i></button>
     <button class="tablinks tablinks-right" onclick="location.href='/logout'" id="logout_button">Logout</button>
 <!--#endif-->
     <div class="GPA gpa_custom-select custom-select tablinks-right" style="width: 270px">
--- a/public/home.html
+++ b/public/home.html
@@ -73,19 +73,27 @@
     </label>
   </div>
 
-  <div class="tab">
-    <button class="tablinks" onclick="openTab(event, 'grades')" id="default_open">Grades</button>
-    <button class="tablinks" onclick="openTab(event, 'schedule')">Schedule</button>
-    <button class="tablinks" onclick="openTab(event, 'recent')">Attendance</button>
-    <!--<button class="tablinks" onclick="openTab(event, 'clock')">Clock</button>-->
-    <!-- TODO: post-covid, setting lunch should be migrated to settings -->
-    <!--<button class="tablinks" onclick="openTab(event, 'calendar-tab')">Calendar</button>-->
-    <button class="tablinks" onclick="openTab(event, 'reports')" id="reports_open">Reports</button>
-    <button class="tablinks" onclick="openTab(event, 'information')" id="information_open">Info</button>
-<!--#ifndef lite-->
-    <button class="tablinks tablinks-right" onclick="location.href='/logout'">Logout</button>
-<!--#endif-->
-    <div class="GPA gpa_custom-select custom-select" style="width:270px;">
+  <!-- SIDENAV STUFF -->
+  <div id="sidenav" class="sidenav">
+    <button onclick="closeSideNav()" class="tablinks">
+      <i class="fa fa-times" style="position: absolute; top: 20px; left: 20px;"></i>
+      Close
+    </button>
+    <button class="tablinks screensize-show-2" onclick="openTab(event, 'grades');closeSideNav()">Grades</button>
+    <button class="tablinks screensize-show-2" onclick="openTab(event, 'recent');closeSideNav()">Recent Activity</button>
+    <button class="tablinks screensize-show-2" onclick="openTab(event, 'schedule');closeSideNav()">Schedule</button>
+    <button class="tablinks screensize-show-2" onclick="openTab(event, 'clock');closeSideNav()">Clock</button>
+    <!--<button class="tablinks" onclick="openTab(event, 'calendar-tab');closeSideNav()">Calendar</button>-->
+    <button class="tablinks screensize-show-2" onclick="openTab(event, 'reports');closeSideNav()" id="reports_open">Reports</button>
+    <div class="GPA tableData_custom-select custom-select", id="navbar_tableData_div">
+      <select id="tableData_select">
+        <option value="0">Current Year</option>
+        <option value="0">Current Year</option>
+        <option value="import">Import Data...</option>
+      </select>
+    </div>
+    <button class="tablinks tablinks-right" onclick="GPAType();">GPA Type</button>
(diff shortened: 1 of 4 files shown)


andOrlando:
~~merged the branch from info-tab to both debug the dynamic screen resize breakpoints and because now there's actually enough space to fit it in nicely. This will also close #76 partially but people should also update the info tab as they see fit.~~

psvenk:
This pull request currently depends on #157 because it adds the info tab to the hamburger menu. I have added the current contents of #157 to this PR, but this can be replaced with the final info tab (using a rebase) when #157 is done.

andOrlando:
- move hamburger to left
- fix close button
- remove weird color merge in sidenav

psvenk:
@andOrlando Looks good. The only remaining blocker is that there is no visual indication of the currently selected tab (as there is in the status quo). We also thought that it would be nice to have an animation when the sidenav changes size, but this can be addressed in a separate pull request (#197).

public/home.html
```diff
@@ -64,18 +65,29 @@
     <h4 id="title" onclick="openTab(event, 'grades')">Aspine</h4>
   </div>
 
+  <!-- SIDENAV STUFF -->
+  <div id="sidenav" class="sidenav">
+    <!-- switch-exempt makes the switch out of navbar pass over this -->
+    <button onclick="closeSideNav()" class="tablinks switch-exempt" style="margin: 0px;width: 270px; height: 44px; background-color: var(--lightgray1)">
+      <i class="fa fa-times fa-2x" style="position: absolute; top: 18px; left: 18px; font-size: 1.3em; color: #000""></i>
```
psvenk: Remove the double closing quote - while it may still work in some browsers, I think it's still a syntax error and it confuses some syntax highlighters
```suggestion
      <i class="fa fa-times fa-2x" style="position: absolute; top: 18px; left: 18px; font-size: 1.3em; color: #000"></i>
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
 
 }
 
+let initialize_resize_hamburger = function() {
+
+  //total width of all items in tab minus hamburger
```
psvenk: Add spacing and capitalization in comments. In this case it would be the following, but you would need to make the change on all of your comments.
```suggestion
  // Total width of all items in tab minus hamburger
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+  //gets all non-tablinks-right elements and adds their width to total_width  and also left_width
+  $('.tab .tablinks:not(.tablinks-right)').outerWidth(function(_i, w) {total_width += w; left_width += w;});
+
+  //gets all tablinks-right elements and adds their width to total_width
+  $('.tab .tablinks-right').outerWidth(function(_i, w) {total_width += w;})
+
+  const switch_left_items = function() {
+    //checks if left items are in sidenav
+    const in_sidenav = $('.tab .tablinks:not(.tablinks-right, .switch-exempt)').length == 0
```
psvenk: Triple equals
```suggestion
    const in_sidenav = $('.tab .tablinks:not(.tablinks-right, .switch-exempt)').length === 0
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
 
 }
 
+let initialize_resize_hamburger = function() {
+
+  //total width of all items in tab minus hamburger
+  let total_width = -44.25;
+
+  //width of all items to the left (because they get removed last) plus logout button and hamburger widths
```
psvenk: Don't go over 80 characters per line

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+    const in_sidenav = $('.tab .gpa_custom-select').length == 0
+
+    if (in_sidenav) {
+
+      //adds right items to tab in reverse
+      const children = $('#sidenav .tablinks-right:not(#logout_button)')
+      for (i=0; i<children.length; i++) {
+        $(".tab").append(children[children.length-1-i])
+      }
```
psvenk: Adding the elements in reverse can be done in a more intuitive, more idiomatic, and less error-prone way than a C-style `for` loop:
```suggestion
      for (const child of [...children].reverse()) {
        $(".tab").append(child);
      }
```
Here, `[...children]` converts `children` to an array. One can then iterate over it in reverse order using a `for`…`of` loop and `Array.reverse()`.

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+      //also closes the sidebar
+      closeSideNav();
+    }
+    else {
+
+      //adds right items to sidenav in reverse
+      const children = $('.tab .tablinks-right:not(#logout_button)')
+      for (i=0; i<children.length; i++) {
+        $("#sidenav").append(children[children.length-1-i])
+      }
```
psvenk: As before, this can be done better with a ```for```…```of``` loop:
```suggestion
      for (const child of [...children].reverse()) {
        $("#sidenav").append(child)
      }
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+      //adds right items to tab in reverse
+      const children = $('#sidenav .tablinks-right:not(#logout_button)')
+      for (i=0; i<children.length; i++) {
+        $(".tab").append(children[children.length-1-i])
+      }
+
+      //also closes the sidebar
+      closeSideNav();
+    }
+    else {
```
psvenk: Put `else` on the same line as the closing brace of the `if`:
```suggestion
    } else {
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
 
 }
 
+let initialize_resize_hamburger = function() {
```
psvenk: Avoid the `let` … `= function()` pattern for top-level functions (because it leads to inconsistent behavior w.r.t. semicolon placement and hoisting):
```suggestion
function initialize_resize_hamburger() {
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+  //total width of all items in tab minus hamburger
+  let total_width = -44.25;
+
+  //width of all items to the left (because they get removed last) plus logout button and hamburger widths
+  //44.25 is hamburger width, has to be hardcoded becuase it might be display: none;'d
+  let left_width = 44.25 + $('#logout_button').outerWidth()
+
+  //gets all non-tablinks-right elements and adds their width to total_width  and also left_width
+  $('.tab .tablinks:not(.tablinks-right)').outerWidth(function(_i, w) {total_width += w; left_width += w;});
```
psvenk: Use arrow function syntax and don't make overlong (>80 chars) lines; also, there is no need to give a name to the throwaway argument (`_` is fine):
```suggestion
  $('.tab .tablinks:not(.tablinks-right)').outerWidth((_, w) => {
    total_width += w;
    left_width += w;
  });
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+  //gets all tablinks-right elements and adds their width to total_width
+  $('.tab .tablinks-right').outerWidth(function(_i, w) {total_width += w;})
```
psvenk: Use arrow function syntax:
```suggestion
  $('.tab .tablinks-right').outerWidth((_, w) => {total_width += w;})
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+  const switch_left_items = function() {
```
psvenk: Use arrow function syntax for all local functions:
```suggestion
  const switch_left_items = () => {
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+  const switch_hamburger = function() {
+    //if hamburger is hidden
+    const is_hidden = $('#hamburger_button').hasClass("hide")
+    
+    //hides or unhides the hamburger
+    if (is_hidden) {
+      $('#hamburger_button').removeClass("hide")
+    } else {
+      $('#hamburger_button').addClass("hide")
+    }
```
psvenk: This is needlessly verbose and can be shortened to this:
```suggestion
    if ($('#hamburger_button').hasClass("hide")) {
      $('#hamburger_button').removeClass("hide")
    } else {
      $('#hamburger_button').addClass("hide")
    }
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+    switch_right_items();
+    switch_left_items();
+    switch_hamburger();
+  }
+
+  /* ADDS THE EVENT LISTENER */
+  window.addEventListener("resize", function () {
+
+    //updates navbar_state
+    update_navbar_state()
```
psvenk: Add a semicolon
```suggestion
    update_navbar_state();
```

public/js/extraFunctions.js
```diff
@@ -654,6 +654,143 @@ function tableData_closeAllSelect(elmnt) {
...
+
+    //if they're different, updates the tab and sidenav
+    if (old_navbar_state != navbar_state) {
```
psvenk: Triple equals
```suggestion
    if (old_navbar_state !== navbar_state) {
```

public/js/home.js
```diff
@@ -1130,6 +1132,26 @@ function openTab(evt, tab_name) {
     recentAttendance.redraw();
 }
 
+function openSideNav() {
+    let sidenav = document.getElementById("sidenav");
+    if (sidenav.clientWidth == 270) sidenav.style.width = "0px";
+    else sidenav.style.width = "270px";
```
psvenk: Triple equals and use ternary operator (instead of if/else without braces)
```suggestion
    sidenav.style.width = sidenav.clientWidth === 270 ? "0px" : "270px";
```

public/js/home.js
```diff
@@ -1130,6 +1132,26 @@ function openTab(evt, tab_name) {
...
+    const mainSection = $("#sidenav-overlay");
+    mainSection.css("visibility", "visible");
+    $(mainSection).fadeIn("slow");
+}
+
+function closeSideNav() {
+    document.getElementById("sidenav").style.width = "0px";
+    const mainSection = $(".sidenav-overlay");
+    $(mainSection).fadeOut("slow");
+    // mainSection.css("visibility", "hidden");
```
psvenk: Don't keep dead code in comments; we can see any relevant dead code via git.
```suggestion
```

psvenk (changes requested):
Here are some code style changes. In general, a couple more things:
- Be more consistent in your usage of single vs. double quotes. The convention for this codebase is to use double quotes (at least in the frontend), so it would be good to go with that.
- It would be good to avoid jQuery for new code, but my view is that this code is already written and so it is not as much of a priority here.

psvenk (approved):

PR #169, opened 2020-11-12
Changed Aspine header onclick behavior to open grades instead of clock

### LINKED ISSUE #168: Clicking on "Aspine" goes to clock instead of home page
Clicking on the header "Aspine" just opens a full-size clock; this behavior is counterintuitive and should be amended to open the home page ("Grades").

Daniel-Benzion:
Fix for issue #168 ("Clicking on 'Aspine' goes to clock instead of home page").

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -55,7 +55,7 @@
     </strong></p>
   </noscript>
 
-  <div id="header" onclick="openTab(event, 'clock')">
+  <div id="header" onclick="openTab(event, 'grades')">
     <div class="logo">
       <canvas id="small_clock" width="400" height="400"></canvas>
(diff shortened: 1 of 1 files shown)


MohanMaker:
Thank you for the pull request! I see two small issues. Now when you click on the clock in the header, it no longer brings you to the clock page. We want to retain that ability while also adding the ability to click on other parts of the header to bring you back the home page. Additionally, when you click the header, you are brought to the Grades page, but the words "Grades" isn't greyed indicating that you are on that tab. Thanks!

psvenk:
@Daniel-Benzion Thank you for the pull request. Because @MohanMaker fixed the same issue in #177 while also addressing the action items on this pull request, I am going to close this now.

PR #181, opened 2020-11-20
Fix handling of failed logins

psvenk:
Avoid throwing an error on the backend when a user types an incorrect username or password.

Known issue: the user is directed to import data.


psvenk:
Obsolete with the new API; see #226 for further developments.

PR #182, opened 2020-11-21
Add support for previous year grades

### LINKED ISSUE #99: Get grades from previous year
Similar to #73, but selecting "Previous year" in the year dropdown rather than selecting a term in the term dropdown. If this is implemented after #74, allow exporting data from the previous year.

psvenk:
This is definitely not yet ready to be merged, but it would close #99. These are commits that I originally made in June (as `v2.3.0-21-gc884aa3`) and rebased today on the latest `master` (as `v2.5.1-3-g7f941f7`).

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -91,6 +91,7 @@
       <select id="tableData_select">
         <option value="0">Current Year</option>
         <option value="0">Current Year</option>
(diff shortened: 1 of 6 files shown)


psvenk:
We have resolved to wait for the new API; see #99 for further development.

Issue #167, opened 2020-11-09
Display more information for assignments without statistics


psvenk:
Currently, certain assignment information such as date assigned, date due, and assignment feedback are only visible for assignments that have statistics. We should allow the user to view this sort of assignment information even when there are no assignment statistics available (in which case there should be no box plot).

psvenk:
Fixed in #297.

PR #183, opened 2020-11-22
Change stats symbol

### LINKED ISSUE #185: Changes to icons
- Change toolbox icon for test corrections to something more indicative of its purpose
- Change color of i and hammer buttons from black to something else

jadebuckwalter:
Replace the "i" icon with a bar graph symbol.


jadebuckwalter:
I also changed the color of the toolkit from green to black as we discussed in last week's meeting.

psvenk:
Thank you @jadebuckwalter for the change. I think it would be better to use the "i" icon next to each assignment while using the bar graph icon in the header, as this would coordinate better with the symbolism of the toolbox and hammer (the toolbox better indicates the action while the hammer communicates something auxiliary or subordinate), removes the need for the inverted icon (with the regular icon looking more harmonious with the toolbox icon), and also is consistent with #167 once implemented. Your thoughts?

jadebuckwalter:
@psvenk Yes, definitely! The inverted bar graph looks a little strange anyway once it’s shrunken down to that size.

psvenk (approved):

PR #195, opened 2020-12-09
Dark Mode

### LINKED ISSUE #146: Dark mode/theme
I think it would be cool to have a dark theme, seeing as it's all the rage these days. It could be especially helpful on mobile. This could potentially be a pretty massive CSS undertaking, so it's definitely not urgent, but something to think about.

jadebuckwalter:
Create a switch that when toggled, switches Aspine to Dark Mode.


jadebuckwalter:
Some notes about Dark Mode:

Aspine switches to Dark Mode by changing the values of color variables in the CSS files, either by detecting the color scheme that the device is already using or by toggling a switch in the upper right-hand corner. If the device is already in Dark Mode, Aspine will automatically change to Dark Mode upon loading. However, the toggle works whether or not the device is already in Dark Mode.

While working on this, I found many instances of text that did not have a color variable assigned to it in the CSS file. This is a problem because the text will not change color when Dark Mode is toggled, making it hard to read. I am fairly confident that I fixed all of these instances, but if there are any remaining (except for the schedule - more on that below), please let me know.

The Dark Mode toggle does not quite work with the Schedule tab (yet). The colors for each class period are stored in an array in the JS file rather than in the CSS file, so they are only read when the page initially loads. This is not a huge problem, because the text can still be read when hovered over, but it is far from ideal and something I am still working on.

I also changed the bar graph icon I had previously added to a similar icon from Material Icons. This makes it easier to change the color of the icon.

tektaxi:
Dark Mode Bugs:

- [x] 1. Aspine Header and small clock can't be clicked on to take you to their respective pages. (grades, clock). The pointer also doesn't change on hover, so CSS Isn't detecting it.

- [x] 2. Schedule Tab doesn't switch when the mode switches
- [x] 3. Hide dark mode toggle on small screens.
- [x] 4. Store user preference of color scheme in localstorage
- [x] 5. Dark mode toggle doesn't load in checked position when system dark mode is loaded.
- [x] 6. Test corrections pane in dark mode - can not see the input box
- [x] 7. Borders around dropdown menus are green. Switch to gray as we talked about.
Add any more issues here

jadebuckwalter:
Didn't have enough space to write this in the commit message, but along with changing the border color of the input box in dark mode to make it visible, I changed the color of the label text to a lighter shade of gray in dark mode to increase the contrast.

public/js/home.js
```diff
@@ -52,6 +52,101 @@ window.addEventListener("click", function(event) {
...
+
+        r.style.setProperty('--blue1', '#9ebcea');
+        r.style.setProperty('--blue3', '#1d68cd');
+
+        localStorage.setItem("color-scheme", "light");
+        dark = false;
+    }
+ }
+
+function getMode() {
```
notrodes: Delete, don't need it, see later comments for why.

public/js/home.js
```diff
@@ -969,7 +1064,7 @@ function scheduleCallback(response) {
     periods = periods.filter(Boolean).map(x => "Period: " + x);
 
     let colors = ["#63C082", "#72C68E", "#82CC9B", "#91D2A7", "#A1D9B4", "#B1DFC0", "#C0E5CD", "#D0ECD9"];
-
+    let darkModeColors = ["#00290e", "#003d14", "#00551d", "#007a29", "#008f30", "#00a336", "#00b83d", "#00cc44"];
```
notrodes: delete, use css variables in the colors list

public/js/home.js
```diff
@@ -987,14 +1082,28 @@ function scheduleCallback(response) {
 
     for (let i = 0; i < periods.length; i++) {
         if (currentTableData.schedule.black[i]) {
-            currentTableData.schedule.black[i].period = periods[i] ? periods[i] + "<br>" + times[i] : "Extra";
-            currentTableData.schedule.black[i].class = currentTableData.schedule.black[i].name + "<br>" + currentTableData.schedule.black[i].teacher;
-            currentTableData.schedule.black[i].color = colors[i] ? colors[i] : colors[colors.length - 1];
+            if (getMode()) {
```
notrodes: delete if, both branches now use colors[] because they use variables

public/js/home.js
```diff
@@ -987,14 +1082,28 @@ function scheduleCallback(response) {
...
+                currentTableData.schedule.black[i].period = periods[i] ? periods[i] + "<br>" + times[i] : "Extra";
+                currentTableData.schedule.black[i].class = currentTableData.schedule.black[i].name + "<br>" + currentTableData.schedule.black[i].teacher;
+                currentTableData.schedule.black[i].color = darkModeColors[i] ? darkModeColors[i] : darkModeColors[darkModeColors.length - 1];
+            }
         }
         if (currentTableData.schedule.silver[i]) {
-            currentTableData.schedule.silver[i].period = periods[i] ? periods[i] + "<br>" + times[i] : "Extra";
-            currentTableData.schedule.silver[i].class = currentTableData.schedule.silver[i].name + "<br>" + currentTableData.schedule.silver[i].teacher;
-            currentTableData.schedule.silver[i].color = colors[colors.length - 1 - i] ? colors[colors.length - 1 - i] : colors[0];
+            if (getMode()) {
```
notrodes: same as above

notrodes (changes requested):

public/js/home.js
```diff
@@ -52,6 +52,101 @@ window.addEventListener("click", function(event) {
...
+
+        r.style.setProperty('--black', '#000000');
+
+        r.style.setProperty('--blue1', '#9ebcea');
+        r.style.setProperty('--blue3', '#1d68cd');
+
+        localStorage.setItem("color-scheme", "light");
+        dark = false;
+    }
+ }
```
psvenk: This should be substantially reworked. The status quo is that the color definitions are repeated between `home.css` and `home.js`. Instead, `:root` and `@media (prefers-color-scheme: dark) { :root { ... } }` can be changed to be `body` and `body.dark`, and we can set/unset the `dark` class in JS according to `window.matchMedia('(prefers-color-scheme: dark)').matches` and the setting in `localStorage` (if any). Then, this function would simply need to toggle the `dark` class on the `body` element.

public/home.html
```diff
@@ -26,6 +26,7 @@
   <script type="text/javascript" src="js/extraFunctions.js"></script>
   <script type="text/javascript" src="js/buttonFunctions.js"></script>
   <link rel="stylesheet" type="text/css" href="fonts/fontawesome/css/all.min.css">
+  <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
```
psvenk: This should be included via `node_modules` and exposed to the frontend via `serve.js`. I can take care of this; I am just noting this here as a to-do item.

psvenk (changes requested):

psvenk (approved):
Looks good to me.

tektaxi (approved):
Now that we've cleaned it up it looks great to me! Ready to merge I think.

public/js/home.js
```diff
@@ -52,6 +52,54 @@ window.addEventListener("click", function(event) {
...
+        localStorage.setItem("color-scheme", "light");
+        dark = false;
+    }
+    // Otherwise, toggle back to dark mode
+    else {
+        document.body.classList.toggle("dark");
+        localStorage.setItem("color-scheme", "dark");
+        dark = true;
+    }
+ }
```
psvenk: Simplify code by removing unused variables and needlessly verbose comments
```suggestion
// Toggle between dark and light mode
function darkMode() {
    document.body.classList.toggle("dark");
    localStorage.setItem("color-scheme",
        document.querySelector("body").classList.contains("dark") ?
            'dark' : 'light');
}
```

public/js/home.js
```diff
@@ -52,6 +52,54 @@ window.addEventListener("click", function(event) {
...
+    }
+    else {
+        currentMode = "light";
+    }
+    if (localStorage.getItem("color-scheme") === null) {
+        var preferredMode = currentMode;
+    }
+    else {
+        var preferredMode = localStorage.getItem('color-scheme');
+    }
```
psvenk: This can be greatly simplified (and usage of `var` should be avoided):
```suggestion
    const preferredMode = localStorage.getItem("color-scheme") || currentMode;
```

public/js/home.js
```diff
@@ -52,6 +52,54 @@ window.addEventListener("click", function(event) {
...
+    // loads current os mode and sets slider accordingly
+    const slider = document.getElementById("dark-check");
+    let currentMode = "";
+    if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
+        document.body.classList.toggle("dark");
+        currentMode = "dark";
+        // sets the slider to match the actual mode
+        slider.checked = true;
+    }
+    else {
```
psvenk: `else` should be on the same line as the closing brace
```suggestion
    } else {
```

psvenk (changes requested):
Some more code style changes:

public/js/home.js
```diff
@@ -52,6 +52,54 @@ window.addEventListener("click", function(event) {
...
+        var preferredMode = localStorage.getItem('color-scheme');
+    }
+    // sets mode to preferred if it's different than the current one
+    if (preferredMode !== currentMode) {
+        slider.click();
+        currentMode = preferredMode;
+    }
+    return currentMode;
+};
```
notrodes:
```suggestion
// detect color scheme and set slider
window.addEventListener("load", function loadMode() {
    // loads current os mode and sets slider accordingly
    const slider = document.getElementById("dark-check");
    const preferredMode = localStorage.getItem('color-scheme');
    if (preferredMode ? preferredMode === 'dark' :
            window.matchMedia("(prefers-color-scheme: dark)").matches) {
        document.body.classList.toggle("dark");
        // sets the slider to match the actual mode
        slider.checked = true;
    }
});
```
Modern code

notrodes (changes requested):

public/js/home.js
psvenk: With clearer / more idiomatic indentation:
```suggestion
// detect color scheme and set slider
window.addEventListener("load", function loadMode() {
    // loads current os mode and sets slider accordingly
    const slider = document.getElementById("dark-check");
    const preferredMode = localStorage.getItem('color-scheme');
    if (preferredMode ? preferredMode === 'dark' :
            window.matchMedia("(prefers-color-scheme: dark)").matches) {
        document.body.classList.toggle("dark");
        // sets the slider to match the actual mode
        slider.checked = true;
    }
});
```

psvenk (approved):

notrodes (approved):

PR #213, opened 2020-12-16
Dark Mode Updates

### LINKED ISSUE #210: Updates to dark mode
There are a few problems with dark mode (screenshots to come) that we should fix soon-ish.

- [x] Gray background in Grades and Clock tab is jarring (make it a darker shade)

- [x] Black/silver slider changes color to white and silver (it should stay invariant in dark mode)

- [x] Colors in Reports tab are not great

jadebuckwalter:
Make the background color on info tab and clock darker in dark mode

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -51,8 +51,6 @@ body {
   /* Different variable for gray subtitles that turn to white for dark mode */
   --gray-to-white: #666666;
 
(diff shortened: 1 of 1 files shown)


MohanMaker (approved):
Just tested and it looks good to me

PR #215, opened 2020-12-16
tweak table margins

MohanMaker:
Extremely minor change to make the margins around the grades table look more consistent.


jadebuckwalter (approved):
Looks good to me!

Issue #54, opened 2019-04-14
Add support for full transcript viewing


Ruborcalor:

psvenk:
Description from #100 (now closed as duplicate):
> Aspen includes a "transcript" feature under the "My Info" tab which displays a record of the student's final grades in all classes for all school years. It would be nice if Aspine could display this information in addition to the detailed information for the current year (and previous year(s) after #99 and/or #74 is/are implemented)

psvenk:
We should also have a way of showing the transcript grade for a class, in cases where it differs from the term average.

PR #220, opened 2020-12-19
Migrate to REST API

### LINKED ISSUE #90: Add option to load more reports in Reports tab
Currently, only the 5 most recent reports are shown. I am thinking that we could have a button "Show more…" which sends a request to the server to scrape Aspine and fetch the second/third/nth page of the reports (or determine that there are no more reports to return), so that initial loading time of the Reports tab would be unaffected.

### LINKED ISSUE #113: Some classes show the wrong assignments when switching terms
When I switch from "Current Quarter" (Q3) to Q2, one of my classes shows assignments from the second page of Q2 assignments and the second page of Q3 assignments, instead of showing the first and second page of Q2 assignments. I suspect that this is because Aspen "remembers" that it is on the second page (as it needed to scrape the second page of the "current" quarter after the first page) and gives the second page of Q3 assignments. Aspine's scraper also switches terms instead of trying to go to the next page for some reason or fills in "current quarter" data because it does not see any second page.

I am guessing that this bug can be fixed by having the scraper go back to the first page once done with scraping a class for some quarter.

### LINKED ISSUE #163: Close Aspen sessions when we are done with them
The ICTS noticed increased sessions for student Aspen login, see if there is a way to close session on logout from Aspine.

### LINKED ISSUE #170: Grades page only displays first page of grades as shown in Aspen
Aspine is only displaying the most recent 25 assignments for each class, and though this does not affect the calculation of the overall grade, it affects the grade calculation for each category. 

This may be because Aspine is only accessing the first page of grades displayed in Aspen for each class. Aspen displays assignment grades in pages of 25 assignments, so we need to find a way to import data beyond the first page so that the assignment grades reflect the student's overall grade.

### LINKED ISSUE #184: Migrate to Aspen REST API (via mobile site)
We have been using the old desktop website as a scrape source but the mobile site exposes a real api.

psvenk:
Closes #184, #90, #163, #113, and #170. This pull request migrates from scraping the Aspen website to making use of Aspen's REST API, fixing many scraping bugs in the process.

Some notes about testing the code: some of the build framework had to be changed to accommodate a new TypeScript scraper, but you should just be able to run `npm start` as usual. Please remember to run `npm install` to load necessary dependencies after checking out the `rest` branch.

Known issues / to-do list:
- [x] The frontend code for the "Reports" tab currently expects every file to have a distinct name from the others. For the time being, there is a workaround in the backend to make sure that file names are unique (see 2d0ee1b40f - "Fix bugs with Reports tab"), but this is hardly ideal.
  + Maybe write a function to guess the dates/quarters?
- [x] ~`restart-serve.sh` (for the production server) as well as `README.md` and `start.bat` need to be updated to run `npm start insecure` or to run `tsc` directly.~
- [x] Statistics currently do not work because the frontend expects access to a session ID and an Apache token (while the new scraper handles Aspen sessions on its own). We need a new function in the scraper to fetch statistics for an assignment which handles session creation and destruction; this will probably borrow some code from the existing `scrape.js` implementation because it does not seem that the REST API exposes the assignment statistics.
- [x] Most (eventually all) of `scrape.js` is now dead code and should be removed.
- [x] The new code seems to be significantly slower; we should investigate this.
- [x] The schedule endpoint throws an "Invalid login" error a significant portion of the time; investigate this.
- [x] Recent attendance needs to be implemented in the new scraper.


psvenk:
> - The new code seems to be significantly slower; we should investigate this.

Fixed in 9378b0e705.

psvenk:
> - Recent attendance needs to be implemented in the new scraper.

Fixed in d22a118da9. I also added support for "grades posted" events, so it may be a good idea to (partially) revert #166 and show these events (even if actual recent activity is not shown). We could also show this information in some other format.

psvenk:
> I also added support for "grades posted" events

I have reversed this in 6974ed6084 because reviving the "recent activity" would entail bringing back an area that would be blank most of the time and making the "grades posted" events easy to miss, defeating the purpose. I have made a separate issue (#232) for displaying this information.

psvenk:
> - `restart-serve.sh` (for the production server) as well as `README.md` and `start.bat` need to be updated to run `npm start insecure` or to run `tsc` directly.

Fixed in #233.

kdk1616:
We should enable loading the previous year using the dropdown menu and the new API (for other years after that you can import JSON/Use transcript when that's ready)

psvenk:
@kdk1616 I don't see how this is related directly to the new API. We already have #99 for tracking progress on getting grades from the previous year, which will be addressed once the new API code is done. As for the transcript, that is tracked separately in #54.

psvenk:
> - Statistics currently do not work because the frontend expects access to a session ID and an Apache token (while the new scraper handles Aspen sessions on its own). We need a new function in the scraper to fetch statistics for an assignment which handles session creation and destruction; this will probably borrow some code from the existing `scrape.js` implementation because it does not seem that the REST API exposes the assignment statistics.

Fixed in f2eeaca095, 30009cf480, and 8292df2802.

> - Most (eventually all) of `scrape.js` is now dead code and should be removed.

Fixed in b804953808.

psvenk:
> - The frontend code for the "Reports" tab currently expects every file to have a distinct name from the others. For the time being, there is a workaround in the backend to make sure that file names are unique (see 2d0ee1b40f - "Fix bugs with Reports tab"), but this is hardly ideal.

Fixed in 7dc386aee5.

psvenk:
I have requested review from @notrodes, @jadebuckwalter, @kdk1616, and @tektaxi. Given the major nature of these changes, I would like to have as many people as possible review this PR before it is merged. The reviews of anyone else from these four people are, of course, also welcomed and appreciated.

jadebuckwalter (commented):
For the most part, this looks good! However, there are a few minor fixes (which may or may not be worth implementing):

The assignments for a given class are listed from oldest to newest (oldest at the top, newest at the bottom) rather than from newest to oldest, as they are in Aspen. This is not a major issue; it's just a bit of an inconvenience.

The dropdown in the Reports tab extends very far down the page. Though it's great that the new API can scrape all of the reports rather than just the most recent ones, I'm wondering if this is the best front-end design. It might be better if the dropdown shows a certain number of recent reports with an option to show older reports rather than showing them all at once; after all, it's not too often that people go back and look at their middle school report cards. This is a front-end issue and has nothing to do with the API itself, so it could be something that we fix later on.

tektaxi:
@jadebuckwalter I agree that we should definitely fix the assignments issue. I feel like that would be a big help and it's much more convenient to have recent assignments first

psvenk:
@jadebuckwalter @tektaxi I fixed the assignments issue in 9e12fb2b03. As for the reports issue, I think that would require some fairly substantial changes to the dropdown code, so I am in favor of moving it to a separate issue.

notrodes (approved):

jadebuckwalter (approved):
Looks good!

tektaxi (approved):
Works well!

PR #217, opened 2020-12-17
Update district logo display

psvenk:
Update the way that the district logo, if present, displays on the login page.

Screenshots, grouped by viewport width:

<details>
<summary>425px</summary>

| Before | After |
| -- | -- |
| 425-before | 425-after |

</details>
<details>
<summary>720px</summary>

| Before | After |
| -- | -- |
| 720-before | 720-after |

</details>
<details>
<summary>800px</summary>

| Before | After |
| -- | -- |
| 800-before | 800-after |

</details>
<details>
<summary>1024px</summary>

| Before | After |
| -- | -- |
| 1024-before | 1024-after |

</summary>

PR #221, opened 2020-12-19
Shrink district logo on login page

psvenk:
Due to changes to the login page, a district logo (if present) displays at a very large size — this pull request shrinks it to be less obtrusive. This is a subset of the changes in #217 because it seems that we cannot reach a consensus on that in time for version 2.6.1.


jadebuckwalter (approved):
Looks great!

tektaxi (approved):
Looks good!

kdk1616:
Looks good. Later on we should add the 1024px CSS from #217. But that is not a priority.

kdk1616 (approved):

PR #225, opened 2020-12-23
Fix sidenav highlighting

### LINKED ISSUE #211: Fix highlighting in sidenav
There is no visual indication of the currently selected tab in the sidenav as there is when the sidenav is not open.

jadebuckwalter:
Indicate which tab is currently open by changing the shade of that tab
in the sidenav.

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -1169,7 +1169,7 @@ p#large_clock_period {
   background-color: var(--sidenav-color);
   overflow-x: hidden;
   padding-top: 5px;
-  transition: 0.5s;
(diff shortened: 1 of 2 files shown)


jadebuckwalter:
- [x] Change highlighting in the sidenav so that it matches highlighting in the navbar
- [x] Fix the color scheme in the sidenav so that it matches the rest of Aspine
- [x] Change the background color of a selected tab to a darker grey (or lighter in dark mode)
- [x] Make sidenav an id instead of a class

psvenk (approved):
Looks good to me!

PR #230, opened 2020-12-26
Defer non-critical JS/CSS to improve performance

psvenk:
Restructure parts of `login.html` and `home.html` so that the JS and CSS that is not needed for the first paint is deferred and does not hold up rendering the page.


notrodes:
I dont think that login.js should be defered.

psvenk:
@notrodes I deferred `login.js` because it only contains the "remember me" code. Why do you think this is essential?

notrodes (approved):

PR #233, opened 2020-12-26
use a new library to parse command line args

notrodes:
also some minor cleaning


serve.js
```diff
@@ -16,26 +17,18 @@ const args = require('minimist')(process.argv.slice(2));
...
+    .parse();
 
 // ------------ Web Server -------------------
-const port = 8080;
-
 const app = express();
-app.use(compression({ filter: (..._) => true }));
-app.listen(port, () => console.log(`Aspine listening on port ${port}!`));
+app.use(compression());
+app.listen(program.port, () => console.log(`Aspine listening on port ${program.port}!`));
```
psvenk: @notrodes Why did you remove the custom `filter` function? I had included it because `compression` by default does not compress all traffic and so feeding it a function that always returns `true` causes all traffic to be compressed.

serve.js
```diff
@@ -47,22 +40,19 @@ if(!(args.hasOwnProperty('insecure') || args._.includes("insecure"))) {
...
             // OK, continue
             return next();
         }
         // handle port numbers if you need non defaults
-        // res.redirect('https://' + req.host + req.url); // express 3.x
         res.redirect('https://' + req.hostname + req.url); // express 4.x
-    }
+    }); // at top of routing calls
+
+    http.createServer(app).listen(8090);
```
psvenk:
```suggestion
```
I think this line can be removed because we do not serve HTTP on production.

serve.js
```diff
@@ -47,22 +40,19 @@ if(!(args.hasOwnProperty('insecure') || args._.includes("insecure"))) {
...
+    https.createServer(credentials, app).listen(4430, () =>//443
+        console.log('HTTPS Server running on port 4430') //443
+    );
```
psvenk: Remove outdated comments
```suggestion
    https.createServer(credentials, app).listen(4430, () =>
        console.log('HTTPS Server running on port 4430')
    );
```

serve.js
```diff
@@ -47,22 +40,19 @@ if(!(args.hasOwnProperty('insecure') || args._.includes("insecure"))) {
...
-    function ensureSecure(req, res, next){
-        if(req.secure){
+    app.all('*', (req, res, next) => {
+        if (req.secure) {
             // OK, continue
             return next();
         }
         // handle port numbers if you need non defaults
-        // res.redirect('https://' + req.host + req.url); // express 3.x
         res.redirect('https://' + req.hostname + req.url); // express 4.x
```
psvenk: Remove outdated/unnecessary comment
```suggestion
        res.redirect('https://' + req.hostname + req.url);
```

psvenk (changes requested):
Additional changes which I cannot put as inline comments (because GitHub won't let me comment on lines of code that were not changed):
- `npm install commander` and commit `package.json` and `package-lock.json`.
- `minimist` is no longer needed, so the `require('minimist')` can be removed and we can also `npm uninstall minimist`.
- The variables `privateKey`, `certificate`, and `ca` are not needed and can be folded into the object literal for `credentials`.
- The version number should come dynamically from `package.json`.

serve.js
```diff
@@ -16,26 +17,18 @@ const args = require('minimist')(process.argv.slice(2));
...
-const port = 8080;
-
 const app = express();
-app.use(compression({ filter: (..._) => true }));
-app.listen(port, () => console.log(`Aspine listening on port ${port}!`));
+app.use(compression());
+app.listen(program.port, () => console.log(`Aspine listening on port ${program.port}!`));
 
-if(!(args.hasOwnProperty('insecure') || args._.includes("insecure"))) {
+if(!program.insecure) {
```
psvenk: Fix spacing
```suggestion
if (!program.insecure) {
```

psvenk: This breaks the usage of `npm start insecure`, meaning that we need to update documentation to use `npm start -- -i` instead (which IMO is not very elegant).

notrodes: no, -i or --insecure, which is normal

serve.js
psvenk: By the way, I think it should be fine to change the function from `(..._) => true` to the more concise `() => true`, but we should double-check that everything is being compressed.

serve.js
psvenk: My main concern is not inelegance but that documentation and start scripts need to be updated to use `npm start -- -i`.

serve.js
notrodes: i checked and the default compresses based on what data type it is being used. I though that would be better than everything being compressed.

psvenk: I guess the default algorithm changed since I did this in #125, because now the difference in amount of data transferred is negligible between full compression and default compression. In that case, `compression()` is indeed probably better.

psvenk (approved):

PR #236, opened 2020-12-28
Remove shebang lines from scrape/serve

psvenk:
Because the recommended way to start the Aspine server is `npm start` and not `./serve.js` (especially in light of the oncoming move to the new REST API scraper), there is no use in having shebang lines in `serve.js` or `scrape.js` (particularly considering that the `#!/bin/sh` shebang is hacky and may confuse some editors, causing them to treat these files as shell scripts).

Also, this commit adds a semicolon to an import in `serve.js`.

### DIFF
--- a/scrape.js
+++ b/scrape.js
@@ -1,7 +1,3 @@
-#!/bin/sh
-':' //; exec "$(command -v nodejs || command -v node)" "$0" "$@"
-// https://unix.stackexchange.com/a/65295
-
 // --------------- Parameters ----------------
 // Multi-Threads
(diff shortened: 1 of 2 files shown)


notrodes:
I swear i said this in jan.

notrodes (approved):
better late than never

psvenk:
@notrodes 
> First, they ignore you. Then they ridicule you. Finally, they emulate you.
> – [J. William Grimes and others](https://quoteinvestigator.com/2017/08/13/stages/)

PR #237, opened 2020-12-28
Update dark mode in real time

### LINKED ISSUE #235: Dark Mode Update Realtime
Aspine only sets enables/disables dark mode based on the devices state when it loads the page

* When the mode of a device is toggled after the site has already been loaded, Aspine doesn't respond
* This is particularly important on mobile where there is no slider
* Add an event listener? so that it can respond

psvenk:
Closes #235. Use an event listener to update dark/light mode on the fly when the user changes their operating system setting and has not expressed a preference in `localStorage`.

### DIFF
--- a/public/js/home.js
+++ b/public/js/home.js
@@ -52,18 +52,30 @@ window.addEventListener("click", function(event) {
         tableData_closeAllSelect();
     }
 });
-// detect color scheme and set slider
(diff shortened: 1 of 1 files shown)


jadebuckwalter (approved):
Looks good. I was actually wondering about this issue earlier while working on dark mode, so I'm glad you were able to fix it.

PR #239, opened 2021-01-03
Use NODE_ENV to detect production/development

psvenk:
Following the changes to the CLI in pull request #233, bring back the `start.sh` script to minimize confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,9 +117,9 @@ downloaded Aspine, and run `npm install`.
   + On Windows, double-click on the script `npminstall.bat`. The `.bat` file
 extension may be invisible depending on your system configuration.
 
-- Start the Aspine server by running `npm start -- -i` in a terminal window (or
+- Start the Aspine server by running `npm start` in a terminal window (or
   Command Prompt on Windows). If you are on a Unix-like system, you can run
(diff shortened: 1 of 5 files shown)


notrodes:
just make a npm run debug script, dont add a file

psvenk:
> just make a npm run debug script, dont add a file

@notrodes What are your thoughts on making use of the `NODE_ENV` environment variable to detect whether to use HTTP or HTTPS by default, with `-s,--secure` and `-i,--insecure` available as overrides? That way, we could use the same `npm start` script (and `restart-serve.sh` would need to be updated to set `NODE_ENV=production`).

notrodes:
> > just make a npm run debug script, dont add a file
> 
> @notrodes What are your thoughts on making use of the `NODE_ENV` environment variable to detect whether to use HTTP or HTTPS by default, with `-s,--secure` and `-i,--insecure` available as overrides? That way, we could use the same `npm start` script (and `restart-serve.sh` would need to be updated to set `NODE_ENV=production`).

Good idea

tektaxi (approved):
Looks good. Glad to have an easy start method again.

PR #240, opened 2021-01-05
Display New Version information from changelog at bottom of Info Page

### LINKED ISSUE #174: Include a way to view the changelog in Aspine
It would be nice if users could see (in a non-intrusive way) what new features have been added to Aspine. We could do this by rendering `CHANGELOG.md` to HTML in the backend and serving that via an endpoint which is loaded upon a button click in the info tab (after #157 is merged; that blocker is why this issue is currently on hold). In the lite version build process, we could include a step that runs a script in JS that does this compilation (we might do well to rewrite the build script entirely in JS in that case).

kdk1616:
* Parses Markdown from CHANGELOG.md and displays at bottom of info page

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -1261,14 +1261,19 @@ p#large_clock_period {
   margin-top: 2em;
 }
 .info-header a {
-  font-size: 100%;
+  font-size: 120%;
+  color: var(--green2);
   text-decoration: none;
+  margin-bottom: 2em;
 }
+
 
 hr {
   margin-top:0px;
 }
-
+#updates ul {
+  margin-bottom: 20px;
+}
 #moreVersions {
     text-align: center;
 }
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -18,7 +18,6 @@
 
 * {
   margin: 0px;
-  padding: 0px;
   box-sizing: border-box;
 }
 
@@ -150,11 +149,6 @@ p {
   line-height: 1.7;
   color: var(--gray-to-white);
   margin: 0px;
-}
-
-ul, li {
-  margin: 0px;
-  list-style-type: none;
 }
 
 body {
@@ -809,6 +803,8 @@ canvas#large_clock {
 #export_modal_content ul {
   margin-left: 1em;
   color: var(--black);
+  list-style-type: none;
+  padding: 0;
 }
 
 #export_modal_content label {
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -1256,3 +1256,20 @@ p#large_clock_period {
   opacity: 0;
   transition: opacity 300ms 0ms, visibility 0ms 300ms;
 }
+
+#updates-container {
+  margin-top: 2em;
+}
+.info-header a {
+  font-size: 100%;
+  text-decoration: none;
+}
+
+hr {
+  margin-top:0px;
+}
+
+#moreVersions {
+    text-align: center;
+}
+
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -1278,3 +1278,7 @@ hr {
     text-align: center;
 }
 
+#moreVersions a{
+  font-size: 120%;
+}
+
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -1274,11 +1274,10 @@ hr {
 #updates ul {
   margin-bottom: 20px;
 }
-#moreVersions {
+#more-versions {
     text-align: center;
 }
 
-#moreVersions a{
+#more-versions a {
   font-size: 120%;
 }
-
--- a/public/home.html
+++ b/public/home.html
@@ -416,8 +416,26 @@
       GPA for each quarter, which when clicked allows you to see your grades and
       assignments for that quarter.
       </p>
-    </div>
+	  
+	  
+	  <style>
+		  
+		  .info-header a {
+			  font-size: 100%;
+			  text-decoration: none;
+			  
+		  }
+		  
+		  
+		  </style>
+		  
+	  
+      <p class="info-text"><span id="updates"></span></p>
+ 	 	
+	   
   </div>
+ </div>
+ 
   <p>Aspine version: <span id="version">(loading&hellip;)</span></p>
   <script>
     window.addEventListener('DOMContentLoaded', () => {
--- a/public/home.html
+++ b/public/home.html
@@ -416,16 +416,11 @@
       GPA for each quarter, which when clicked allows you to see your grades and
       assignments for that quarter.
       </p>
-      <style>
-		    .info-header a {
-			    font-size: 100%;
-			    text-decoration: none;
-		    }
-      </style>
-      <br>
-      <br>
-      <p class="info-text"><span id="updates"></span></p>
-      <center><a href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">&lt&lt See Older Versions of Aspine &gt&gt</a>
+
+
+      <hr>
+      <p class="info-text" id="updates-container"><span id="updates"></span></p>
+      <p id="moreVersions"><a  href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">&lt&lt See Older Versions of Aspine &gt&gt</a></p>
       <br>
       <br>
     </div>
--- a/public/home.html
+++ b/public/home.html
@@ -421,11 +421,9 @@
       <p class="info-text" id="updates-container"><span id="updates"></span></p>
       <p id="more-versions">
         <a href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">
-          &lt&lt See Older Versions of Aspine &gt&gt
+          View changes made in older versions of Aspine&hellip;
         </a>
       </p>
-      <br>
-      <br>
     </div>
   </div>
   <p>Aspine version: <span id="version">(loading&hellip;)</span></p>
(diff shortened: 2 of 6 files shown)


tektaxi (approved):
Looks good, glad to be able to see updates in Aspine now.

public/home.html
```diff
@@ -416,8 +416,20 @@ <h2 class="info-header">GPA Type and Switching Quarters</h2>
       GPA for each quarter, which when clicked allows you to see your grades and
       assignments for that quarter.
       </p>
+      <style>
+		    .info-header a {
+			    font-size: 100%;
+			    text-decoration: none;
+		    }
+      </style>
```
psvenk: This CSS rule should be moved to `home.css`.

public/home.html
```diff
@@ -416,8 +416,20 @@ <h2 class="info-header">GPA Type and Switching Quarters</h2>
...
+      <br>
+      <br>
+      <p class="info-text"><span id="updates"></span></p>
+      <center><a href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">&lt&lt See Older Versions of Aspine &gt&gt</a>
```
psvenk: - It seems that you have an unclosed `<center>` tag
- [The `<center>` tag is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center); you should assign an ID to the `<a>` and use CSS instead.

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
...
+    }
+
+
+});
+
+
+
+
+
+
```
psvenk: I don't think we should have this much whitespace. I would just remove all whitespace between the end of the `for` loop adn teh end of the callback function and leave only one line between the end of the entire AJAX call and the `//#endif` (for some reason GitHub isn't letting me put this as a "suggestion" with a diff).

serve.js
```diff
@@ -236,6 +245,11 @@ app.use((req, res) => {
 });
 
 
+
+
+
+
+
```
psvenk: As before, I would also remove this whitespace, especially because you did not touch this code and it would be good to have a cleaner overall diff.

serve.js
```diff
@@ -141,6 +143,13 @@ app.use('/fonts/material-icons/iconfont', express.static(
 // Endpoint to expose version number to client
 app.get('/version', (req, res) => res.send(version));
 
+
+
+app.get('/updates', async (req, res) => {
+    const changelog = await fs.promises.readFile(__dirname + '/CHANGELOG.md');
+    res.send(marked("" + changelog));
```
psvenk: Instead of using type coercion (adding a string to `changelog` to convert it from a `Buffer` to a string), use [`Buffer.prototype.toString`](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end):
```suggestion
    res.send(marked(changelog.toString()));
```

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
 
 //#ifndef lite
 $.ajax("/version").then(ver => $("#version").text(ver));
+
+$.ajax("/updates").then(upt => {
+	$("#updates").html(upt);
+	document.getElementById("changelog").outerHTML = "<h2 class='info-header'>Version History/What's New:</h2>";
+    var items = document.getElementById("updates").getElementsByTagName("h2");
```
psvenk: Use `const` instead of `var` and [`document.querySelectorAll`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) instead of `getElementById` and `getElementsByTagName`:
```suggestion
    const items = document.querySelectorAll("#updates h2");
```

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
...
+    for(var i = 3; i < items.length; i ++) {
+        items[i].style = "display: none";
+    }
+    items = document.getElementById("updates").getElementsByTagName("ul");
+    for(var i = 3; i < items.length; i ++) {
+        items[i].style = "display: none";
+    }
+
+
+});
```
psvenk: The indentation for this entire callback should be normalized.

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
...
 //#ifndef lite
 $.ajax("/version").then(ver => $("#version").text(ver));
+
+$.ajax("/updates").then(upt => {
+	$("#updates").html(upt);
+	document.getElementById("changelog").outerHTML = "<h2 class='info-header'>Version History/What's New:</h2>";
+    var items = document.getElementById("updates").getElementsByTagName("h2");
+	for (var i = 0; i < items.length; i++) {
+	    items[i].className = "info-header";
+	}
```
psvenk: Instead of using a C-style `for` loop prone to off-by-one errors, this can be done more simply:
```suggestion
	items.forEach(x => { x.className = "info-header"; });
```

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
...
+    //Hide Everything but first and second versions
+    for(var i = 3; i < items.length; i ++) {
+        items[i].style = "display: none";
+    }
```
psvenk: This can also be done more directly and in a way that avoids potential conflict with other styling (just good practice in general):
```suggestion
    items.slice(3).forEach(x => { x.style.setProperty("display", "none") });
```

public/js/home.js
```diff
@@ -1264,6 +1264,31 @@ document.getElementById("default_open").click();
...
+    items = document.getElementById("updates").getElementsByTagName("ul");
+    for(var i = 3; i < items.length; i ++) {
+        items[i].style = "display: none";
+    }
```
psvenk: This can also be simplified:
```suggestion
    const subpoints = document.querySelectorAll("#updates ul");
    subpoints.slice(3).forEach(x => { x.style.setProperty("display", "none") });
```

psvenk (changes requested):
Looks good overall (from reading the diff), though I have not yet tested your code. Here are some code style changes (but take them with a grain of salt because I haven't tested any of my proposed changes).

notrodes:
Stop writing jquery!!

notrodes:
Im fine not rewriting stuff, just stop writing morehow hard is that!?!?!?

notrodes (dismissed):
delete $

public/js/home.js
```diff
@@ -1268,21 +1268,16 @@ $.ajax("/version").then(ver => $("#version").text(ver));
 $.ajax("/updates").then(upt => {
 	$("#updates").html(upt);
 	document.getElementById("changelog").outerHTML = "<h2 class='info-header'>Version History/What's New:</h2>";
-    const items = document.querySelectorAll("#updates h2");
+    var items = [...document.querySelectorAll("#updates h2")];
```
psvenk: @kdk1616 Why did you change `const` to `var` here? We should always avoid `var` in favor of `const` or `let`.

public/js/home.js
```diff
@@ -1268,21 +1268,16 @@ $.ajax("/version").then(ver => $("#version").text(ver));
...
 	items.forEach(x => { x.className = "info-header"; });
+    console.log(items);
```
psvenk: Remove log statement
```suggestion
```

public/js/home.js
```diff
@@ -1268,21 +1268,16 @@ $.ajax("/version").then(ver => $("#version").text(ver));
...
     //Hide Everything but first and second versions
     items.slice(3).forEach(x => { x.style.setProperty("display", "none") });
-    const subpoints = document.querySelectorAll("#updates ul");
-    subpoints.slice(3).forEach(x => { x.style.setProperty("display", "none") });
-
-
+    const subpoints = [...document.querySelectorAll("#updates ul")];
+    console.log(subpoints);
```
psvenk: Remove log statement
```suggestion
```

psvenk (changes requested):

public/css/home.css
```diff
@@ -1256,3 +1256,25 @@ p#large_clock_period {
...
+}
+
+
+hr {
+  margin-top:0px;
+}
+#updates ul {
+  margin-bottom: 20px;
+}
+#moreVersions {
```
psvenk: Use kebab-case for CSS IDs and classes, not camelCase
```suggestion
#more-versions {
```

public/home.html
```diff
@@ -416,8 +416,14 @@ <h2 class="info-header">GPA Type and Switching Quarters</h2>
       GPA for each quarter, which when clicked allows you to see your grades and
       assignments for that quarter.
       </p>
+
+      <hr>
+      <p class="info-text" id="updates-container"><span id="updates"></span></p>
+      <p id="moreVersions"><a  href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">&lt&lt See Older Versions of Aspine &gt&gt</a></p>
```
psvenk: Remove extraneous space between `a` and `href` and prevent overlong lines; use
kebab-case instead of camelCase for CSS ID
```suggestion
      <p id="more-versions">
        <a href="https://github.com/Aspine/aspine/blob/master/CHANGELOG.md">
          &lt&lt See Older Versions of Aspine &gt&gt
        </a>
      </p>
```

psvenk (changes requested):

psvenk (approved):
Looks good to me; I think this is ready to merge pending @notrodes' approval. I noticed that the changelog inclusion does not work in Aspine Lite, but I chose to split that into a separate issue (#250) (a) this is a sufficiently small feature, (b) the link to GitHub still exists on Aspine Lite, and (c) implementing this would require considerable work on the build script.

PR #242, opened 2021-01-06
Fix period numbering under clock

### LINKED ISSUE #199: Period numbering incorrect in clock
When a student has a different number of classes on black days than on silver days (particularly when there is an AM block in one but not the other), the period names do not update when switching between black and silver.

jadebuckwalter:
Write a conditional to check whether a schedule includes a "before
school" class, as well as to align the black and silver day schedules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
 
-[{serve.js,scrape.js,public/js/home.js}]
+[{serve.js,scrape.js,public/js/home.js,public/js/clock.js}]
 indent_style = space
 indent_size = 4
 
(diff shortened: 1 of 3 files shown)


psvenk (changes requested):
This seems to work with my schedule, but it seems a bit hacky. Instead of computing an index using the period number (`Number(default_name.charAt(default_name.length - 1))`), we could just compare the last character of `default_name` with the last character of each entry of `period_names[bs_day]` (`for (const { name } of period_names[bs_day]) { ... }`, for example). If there is a match, then we can assume that this is the correct period and return the name; if no match, then just return the `default_name`. This way, we don't have to special-case "Before School" (AM) classes, we don't need to assume that they all have the words "Before School" in the title, we don't have separate logic for black and silver days, and the code is overall a lot simpler. Your thoughts?

jadebuckwalter:
@psvenk Yes, that's a better idea.

jadebuckwalter:
@psvenk This should work, and it doesn't rely on detecting before school classes.

psvenk (approved):
@jadebuckwalter I made a few changes; do you think this is ready to merge now?

PR #245, opened 2021-01-20
Update readme

jadebuckwalter:
Update the readme file to include updated info on the color schemes for
both light and dark mode.


jadebuckwalter:
Should I also update the "features" section to include some of the new features? I don't think there are any major features that aren't included, so this may not be necessary.

jadebuckwalter:
Also, should I include the various shades of gray in the readme? It may seem a bit unnecessary but it might be helpful to further contextualize the shades of green in dark mode.

README.md
```diff
@@ -19,11 +19,11 @@ If you have an iPhone and prefer to use an app instead of the website, Aspine is
 
 ![CRLS Clock Image](images/crls-clock.png)
 
-* Grade Comparison via boxplot
+* Grade Comparison via Box Plot
 
 ![Example Boxplot Image](images/boxplot.png)
 
-* Export and import your grades
+* Export and Import your Grades
```
psvenk: It is good to normalize the letter case here, but I think it would be better to normalize to sentence case instead of title case. Your thoughts @jadebuckwalter?

README.md
```diff
@@ -171,7 +180,7 @@ Our privacy policy can be viewed at <https://www.cpsd.us/privacy>.
 
 Copyright
 [Aspine contributors](https://github.com/Aspine/aspine/graphs/contributors)
-2020.
+2019-21.
```
psvenk: Minor typography nitpick: use an en dash.
```suggestion
2019&ndash;21.
```

psvenk (commented):
Looks good to me overall, but here are some comments:

README.md
jadebuckwalter: @psvenk Yes, that makes sense, but it also contributes to greater inconsistencies, like "Aesthetic Grade Checking" being in title case while "Export and import your grades" being in sentence case. If that's ok, then I'm all for switching to sentence case.

psvenk: @jadebuckwalter I meant to suggest putting the entire section in sentence case. The reason why I only commented on these lines is that GitHub did not allow me to select lines that you left untouched.

jadebuckwalter: @psvenk I see. Sorry for the confusion.

psvenk (approved):
Looks good to me!

PR #249, opened 2021-01-27
Clock updates

jadebuckwalter:
Hide period name under the clock so it isn't wrong with the new schedule


psvenk:
@jadebuckwalter Thanks for the second commit; I was about to make the same change myself. Now I'll revert the first commit and squash and merge.

jadebuckwalter:
@psvenk Great, thanks. I didn't see this until after I made that commit.

psvenk (approved):

PR #252, opened 2021-02-02
Added some quality of life tooling

notrodes:
Generate self signed certs for tests easily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ db/*.rdb
 usage_log.txt
 .DS_Store
 dist-lite/
-local*
+local.*
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ db/*.rdb
 usage_log.txt
 .DS_Store
 dist-lite/
+local*
--- a/gen-key.sh
+++ b/gen-key.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+#https://www.rosehosting.com/blog/how-to-generate-a-self-signed-ssl-certificate-on-linux/
+openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out local.crt -keyout local.key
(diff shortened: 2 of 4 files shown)


serve.js
```diff
@@ -49,7 +50,7 @@ app.listen(program.port, () =>
     console.log(`Aspine listening on port ${program.port}!`)
 );
 
-if (program.secure) {
+if (!program.insecure) {
```
psvenk: There is no `program.insecure` because of how I implemented the `secure`/`insecure` (secure by default if and only if production). Instead, this can be changed to `if (program.secure || program.dev)` for the expected behavior with a plain `npm start` for development.

.gitignore
```diff
@@ -3,3 +3,4 @@ db/*.rdb
 usage_log.txt
 .DS_Store
 dist-lite/
+local*
```
psvenk:
```suggestion
local.*
```
Avoid accidentally including any filename beginning with "local" in the future

gen-key.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#https://www.rosehosting.com/blog/how-to-generate-a-self-signed-ssl-certificate-on-linux/
```
psvenk: Add space before comment
```suggestion
# https://www.rosehosting.com/blog/how-to-generate-a-self-signed-ssl-certificate-on-linux/
```

psvenk (changes requested):
Here are some suggested style changes. However, this begs the question: is it really necessary to have SSL certificate tooling for development use? There is no real point in using HTTPS when connections are typically not even exposed to a LAN, much less the Internet.

notrodes:
@psvenk I wanted to have this for when I work on the server.js and want to on https, wanted to allow others to do it easily as well.

psvenk (approved):

psvenk (commented):
I approve, but let's hold off on merging until 2.6.2 is put on production.

PR #255, opened 2021-02-06
Fix schedule + clock for semester 2

### LINKED ISSUE #247: Update clock to match second semester schedule
Change the timing of the schedule so that the class name displayed under the clock is correct.

psvenk:
This pull request makes some changes to the backend and the frontend to
adapt to the semester 2 schedule. On the backend, it excludes the Study
Support block from being counted as a class so that it can be included
on the frontend. On the frontend, a few changes have been made:
- Update `schedule.json` to include the new schedule (Monday/Tuesday,
  Wednesday, and Thursday/Friday variations)
- Use the Monday/Tuesday variation of the new schedule instead of the
  old schedule
- Re-enable period name detection

To do:
- [x] Detect the day of the week and use this to select the appropriate
  variation of the schedule as well as black/silver
- [ ] Replace the black/silver slider on the schedule tab with a
  day-of-week picker
- [ ] Change the schedule display to be more in line with the
  progression of the day as well as how the schedule is displayed on
  Aspen (we could draw from `schedule.json` for the order of periods as
  well as times)

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -278,11 +278,12 @@
   </div>
   <div id="schedule" class="tabcontent">
     <!-- Rounded switch -->
-    <label class="switch">
+    <!-- Black/silver switch disabled due to covid-19 schedule -->
+    <!--<label class="switch">
       <input type="checkbox" id="schedule_toggle" onclick="schedule_toggle();">
       <span class="slider round"></span>
       <p id="schedule_title" class="unselectable">Black</p>
-    </label>
+    </label>-->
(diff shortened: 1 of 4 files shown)


psvenk:
It may be a good idea to merge this without the latter two to-do items (and splitting those off into a separate issue), given that those two would entail broad structural changes and that we should merge this into master in time for 2.6.3.

tektaxi (approved):
Works well for me!

jadebuckwalter (approved):

PR #257, opened 2021-02-07
Add day-of-week picker

### LINKED ISSUE #256: Additional changes to schedule tab for semester 2
Split off from the to-do list on #255.

- [x] Replace the black/silver slider on the schedule tab with a day-of-week picker
- [x] Change the schedule display to be more in line with the order of periods in the day as well as how the schedule is displayed on Aspen (we could draw from `schedule.json` for the order of periods as well as times)

jadebuckwalter:
Add a dropdown above the schedule to select the day of the week, and
update the schedule to match that day.

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -285,12 +285,12 @@
       <p id="schedule_title" class="unselectable">Black</p>
     </label>-->
     <select id="day-select" onchange="schedule_toggle(value);">
-      <option id="day">Select Day</option>
-      <option value="mon">Monday (Silver)</option>
-      <option value="tues">Tuesday (Black)</option>
-      <option value="wed">Wednesday</option>
-      <option value="thurs">Thursday (Silver)</option>
-      <option value="fri">Friday (Black)</option>
+      <option value="-1">Select Day</option>
+      <option value="1">Monday (Silver)</option>
+      <option value="2">Tuesday (Black)</option>
+      <option value="3">Wednesday</option>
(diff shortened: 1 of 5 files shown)


jadebuckwalter:
Though the basic functionality has been implemented, this is still a work in progress. This is what remains:

- [x] Default to current day
- [ ] Style the dropdown menu
- [x] Update clock to match dropdown
- [x] Update schedule to match dropdown
- [x] Clean up the code (that's on me)

psvenk (changes requested):
Thanks for the changes. While we should have a day-of-week dropdown, I am still wary of completely refactoring the code to rely on the day of the week and ditching black and silver days altogether, given that we will probably go back to a black/silver-based schedule post-covid. Many parts of the code use conditionals so that we can easily revert to the pre-covid behavior at any time.

Is there any technical reason why this refactoring is necessary, or would it be feasible to continue having the day of the week be used to compute whether it's a black or silver day as well as which schedule from `schedule.json` to use (`covid-mt`, `covid-w`, or `covid-rf`)? The main change would be that instead of directly using `now.getDay()` in those places, we could have a global variable `day` and do something like `day || now.getDay()` so that `day`, if set, overrides `now`. (We would also need some code using something like `new Date().getDay()` to set the dropdown to default to the current day of the week.)

psvenk (approved):
@jadebuckwalter, @tektaxi, and I have agreed that the styling can be addressed in a separate pull request, so this is ready to merge.

PR #262, opened 2021-02-08
Style day of week picker

jadebuckwalter:
Style the dropdown menu so it looks nicer and uniform to other dropdowns

### DIFF
--- a/public/js/extraFunctions.js
+++ b/public/js/extraFunctions.js
@@ -1131,7 +1131,7 @@ let tableData_option_onclick = function() {
 
 let initialize_tableData_dropdown = function() {
   let x, i, j, selElmnt, a, b, c;
-  /* Look for any elements with the class "pdf_custom-select": */
+  /* Look for any elements with the class "tableData_custom-select": */
   x = document.getElementsByClassName("tableData_custom-select");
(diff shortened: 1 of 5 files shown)


public/js/extraFunctions.js
```diff
@@ -1167,6 +1167,50 @@ let initialize_tableData_dropdown = function() {
...
+    b.setAttribute("class", "day_select-items select-hide");
+    for (j = 1; j < selElmnt.length; j++) {
+      /* For each option in the original select element,
+      create a new DIV that will act as an option item: */
+      let bs;
+      if (j === 1 || j === 4) {
+        bs = "silver";
+      } else if (j === 2 || j === 5) {
+        bs = "black";
+      }
```
psvenk: What is this variable `bs` for? It doesn't seem to be used anywhere, and anyway the heavy lifting w.r.t. schedule formatting and black/silver is handled elsewhere.

jadebuckwalter:
Some minor fixes remain:
- [x] If you click out of the dropdown, it should go away
- [x] The arrow should point up if the menu is open
- [x] Change "select day" to the day of the week if it is a weekday

psvenk (approved):
Looks good to me!

PR #268, opened 2021-02-14
Fix bug with AM and PM classes

### LINKED ISSUE #267: AM and PM classes show up multiple times
Either the AM class or the PM class will show up in both spots, instead of one in each.

jadebuckwalter:
Detect difference between AM and PM period names so they show up in the
correct places

### DIFF
--- a/public/js/clock.js
+++ b/public/js/clock.js
@@ -234,15 +234,15 @@ function get_period_name(default_name, day_of_week) {
     }
     // period_names has class names now
     for (const { name, period } of period_names[bs_day]) {
-        if (default_name.slice(-1) === "l") {
-            if (period.slice(0) === default_name.slice(0)) {
-                return name;
-            }
-        } else {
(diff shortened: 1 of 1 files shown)


psvenk (changes requested):
Thank you @jadebuckwalter for the changes. Instead of hard-coding "l" for AM/PM classes, I think it would be a better idea to detect regular periods (01, 02, 03, 04) and fall back to a strict equality check, similar to how it worked before 9d2d9a89af. Maybe we could do a regex search on `period` for `/0\d/` to detect a regular period, and if so, see if the same period number is a substring of the default name. Your thoughts?

jadebuckwalter:
@psvenk Sure, that makes sense.

psvenk (approved):
Looks good to me now. @jadebuckwalter, feel free to squash and merge (maybe using something like my commit message) once you have verified that this works for you.

PR #270, opened 2021-02-16
Show release notes back to last minor version

### LINKED ISSUE #264: Stop hardcoding "past three versions" in changelog code
The changelog in the info tab is hardcoded to show release notes for the three<sup>1</sup> most recent versions of Aspine. Instead of doing this, we should show all release notes going back to the earliest patch level of the current version. For example, on version 2.6.4, the release notes going back to 2.6.0 should be shown.

<sup>1</sup> This was originally 2, but I changed it to 3 in 8f037c891e so that the release notes for 2.6.2 are shown in 2.6.4 (because 2.6.1 was the previous version to go on the production server, and 2.6.2 and 2.6.3 are much more notable than 2.6.4).

psvenk:
Closes #264. Instead of having the changelog code be hard-coded to show the past three versions' release notes, go by minor version numbers instead; that is, display release notes back to the latest minor version release (2.5.0, 2.6.0, etc.).

### DIFF
--- a/public/js/home.js
+++ b/public/js/home.js
@@ -1252,17 +1252,26 @@ $.ajax("/version").then(ver => $("#version").text(ver));
 $.ajax("/updates").then(upt => {
     $("#updates").html(upt);
     document.getElementById("changelog").outerHTML = "<h2 class='info-header'>Version History/What's New:</h2>";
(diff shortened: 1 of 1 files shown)


tektaxi (approved):
Go for it!

PR #271, opened 2021-02-17
Reimplement build script in JS + show changelog in lite version

### LINKED ISSUE #250: Update build script to show changelog in Aspine Lite
#240 only shows updates from the changelog in the info tab in live Aspine and not in Aspine Lite (the purely client-side version generated by `build-lite.sh`). To include this same information in Aspine Lite, we should update the build script so that it renders `CHANGELOG.md` to HTML and includes it in `home.js` using the `#include` pseudo-directive (similarly to how the version number is handled). It may be a good idea to rewrite the build script in Node.js to accomplish this (so that `marked` can be called directly), which is why I split this off into a separate issue.

psvenk:
Fixes #250. The build script has been reimplemented in JavaScript and the changelog is now rendered and included into `home.js` as part of the build process for Aspine Lite, meaning that the "Version History/What's New" section is properly populated in the lite version.


psvenk:
Approval secured from @kdk1616

PR #272, opened 2021-02-23
Add "Incorrect username or password" message

### LINKED ISSUE #226: Give a visual indication of incorrect login info
Currently, when a user enters an incorrect username or password, they are brought to the "Import data" screen, which is not ideal. Instead, we should bring the user back to the login screen and display an "Incorrect username or password" message (see also #46). The backend portion of this issue (distinguishing between an invalid login and no login at all, and bringing the user back to the login screen in the former case) is fixed in #220, so it would probably be wise to wait until that is merged to work on the frontend portion (adding the text to the login page).

psvenk:
Closes #226. Instead of simply sending users back to the login screen when a login fails, alert them to this fact in text above the login form.

### DIFF
--- a/public/css/login.css
+++ b/public/css/login.css
@@ -233,6 +233,11 @@ iframe {
   overflow: hidden;
 }
 
+#login-fail {
+  color: red;
+  display: none;
+}
+
 #bottom-text {
   text-align: left;
   margin: 5px 14px;
--- a/public/js/home.js
+++ b/public/js/home.js
@@ -789,7 +789,7 @@ function responseCallback(response, includedTerms) {
         return;
     }
     if (response.recent.login_fail) {
-        location.href = '/logout';
+        location.href = "/logout?fail=1";
     }
 
     if (response.classes.length === 0) {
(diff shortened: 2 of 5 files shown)


jadebuckwalter (commented):
@psvenk Looks good, although it's a little jarring that it goes to the homepage for a split second before redirecting to the login page. Is there any way to fix that?

psvenk:
@jadebuckwalter Making login validation work without needing to go to the home page would entail communicating with the backend from the login page, so that the backend can talk to Aspen to validate the login. We would furthermore need to do another handshake between the frontend and the backend so that the frontend redirects to the home page or to the "incorrect password" login page depending on the result of this request, and we would need to account for possible race conditions in sending data to the frontend while it is still loading the home page (upon a successful login) and in properly disposing of sessions. This would not only bring a great deal of complexity to the login page (which currently is quite light and only has a little bit of JS) but also make the backend, including `scrape.ts`, much more complex.

Your thoughts?

jadebuckwalter:
@psvenk Ok, makes sense. I wasn't sure how complicated it would be, and based on what you said, it seems like the hassle would outweigh the benefit.

jadebuckwalter (approved):

PR #278, opened 2021-03-11
Handle Aspen outages, improve error handling

### LINKED ISSUE #277: Handle Aspen Errors more gracefully
Right now when Aspen is down, the error message still says "Incorrect username or password." This is strange to see as a user because it seems like you typed something wrong when really it is out of your control. We should handle this differently and give a different and more representative error message to the user.

psvenk:
Closes #277. Distinguish between an Aspen server outage and incorrect login information, creating a new enum in the backend to keep track of error codes.

### DIFF
--- a/public/css/login.css
+++ b/public/css/login.css
@@ -233,7 +233,7 @@ iframe {
   overflow: hidden;
 }
 
-#login-fail {
+#error {
   color: red;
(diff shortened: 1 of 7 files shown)


tektaxi (approved):
looks good!

Issue #23, opened 2019-02-24
Add a settings page.


Ruborcalor:
Add a settings page that allows user to personalize their aspine experience. This would include settings like showing categories grade page, possibly coloring. These settings would ideally save with the username so they wouldn't have to make the changes everytime they log in.

- Opt in/out of notifications from Aspine (after #94)
- Change time zone (after #84)
- Switch between light and dark themes (after #146)

tektaxi:
potentially we could save the user's preferences with localstorage, if we can't store data on the server. That way at least it would be by computer, though not saved to your account.

kdk1616:
Since we're hiding the clock tab, settings could allow user to manually select their lunch. (Again stored in localstorage).

notrodes:
At this point you are going to want to use IndexedDB to save settings for multiple accounts.

psvenk:
> At this point you are going to want to use IndexedDB to save settings for multiple accounts.

@notrodes The plan (at least for now) is to use localStorage so that we do not need a database; however, this can be revisited later with ICTS approval.

PR #283, opened 2021-03-24
Add times to the schedule

### LINKED ISSUE #260: Display period times in schedule
It would be nice to be able to see what time each class is happening at. ex. Period 1 8:35-9:55 etc.

jadebuckwalter:
For each period, display the start and end time in the schedule.

### DIFF
--- a/public/js/clock.js
+++ b/public/js/clock.js
@@ -43,12 +43,14 @@ function schedulesCallback(response) {
     redraw_clock_with_timestamp();
 }
 
-// Takes a time from schedule.json and formats it to hours:minutes
+// Takes a time from schedule.json and formats it as a string
 function formatTime(time) {
-    let minutes = time / 60000;
-    let hours = Math.floor(minutes / 60);
-    minutes = minutes - (hours * 60);
-    return (hours % 12 === 0 ? 12 : hours % 12) + ":" + (minutes < 10 ? "0" + minutes : minutes);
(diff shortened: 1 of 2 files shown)


jadebuckwalter:
It would be great if someone who doesn't have an AM class could test this out.

tektaxi (approved):
Works well for me!

psvenk (changes requested):
This does not work for me. I am working on a fix.

psvenk (changes requested):
I just now refactored the code so that it is more efficient and clean and also works on my schedule, so it looks good to me apart from some stylistic comments: (a) we should display "AM"/"PM" instead of just leaving it implied (we can add a 24-hour time option once #23 is implemented), and (b) an en dash (`–`) should be used between times, not a hyphen (see [Butterick's Practical Typography](https://practicaltypography.com/hyphens-and-dashes.html)).

psvenk:
By the way, the time formatting can be handled by [`Date.prototype.toLocaleTimeString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString). A fix is underway.

jadebuckwalter:
@psvenk Great - that seems a lot easier. I can take care of these fixes if you'd like.

psvenk (approved):
Looks good to me!

jadebuckwalter:
Looks good to me as well.

PR #286, opened 2021-03-24
Use CSS variables in getColor

### LINKED ISSUE #284: Use CSS variables in getColor
When your grade is 100 in a class, and you also have a calculated grade, the calculated grade shows up in black text. When in dark mode, this is a problem because it's really hard to read the black text on the dark grey background. It's a really specific scenario but worth the simple fix.

Screen Shot 2021-03-24 at 4 47 45 PM

jadebuckwalter:
Instead of hard-coding colors, use CSS variables so that grades are
more visible in dark mode.

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -62,6 +62,7 @@ body {
   --purple: #e9a1ff;
 
   --orange: #f6b93b;
+  --orange1: #ff6b00;
 
   --red: #ff0000;
   --red1: #a70000;
@@ -105,6 +106,7 @@ body.dark {
   --purple: #e9a1ff;
 
   --orange: #f6b93b;
+  --orange1: #ff6b00;
 
   --red: #ff0000;
   --red1: #ff8585;
--- a/public/js/buttonFunctions.js
+++ b/public/js/buttonFunctions.js
@@ -12,7 +12,7 @@ let newAssignment = function() {
         "score": 10,
         "max_score": 10,
         "percentage": 100,
-        "color": "green",
+        "color": "var(--green)",
         "synthetic": "true",
       });
     newAssignmentIDCounter++;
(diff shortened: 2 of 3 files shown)


jadebuckwalter:
This fixes the original issue. But in the process I came across `lightColors` - what is that used for, and should those colors get CSS variables as well?

psvenk:
> This fixes the original issue. But in the process I came across `lightColors` - what is that used for, and should those colors get CSS variables as well?

Apparently this was added in a43af604b2 but it seems that it was not used at all then and is not used now, so I think it can be safely deleted (along with `getLightColor`).

psvenk (commented):
This mostly looks good, but I think it would be good to have a separate color (dark orange?) for D-range grades, both for maintaining consistency (every other letter grade has its own color) and because we shouldn't use red for a D when it is still a passing grade.

psvenk (approved):
Should be ready to squash and merge.

psvenk (changes requested):
Actually, I (semi-)retract my approval. I think it may be a good idea to compute the color first (using the variable), given that the color (as returned by `getColor`) will be entered into the JSON export and therefore should be portable.

This would involve re-computing all colors upon switching themes, which might be a little much. @jadebuckwalter Your thoughts?

psvenk:
Or, we could keep using CSS variables within `currentTableData` and just make sure that we do not remove any variables in the future to maintain backwards compatibility. This breaks forwards compatibility, but we never guaranteed forwards compatibility in the first place. It would also be good to change `green` to `var(--green)` inside `newAssignment` in `buttonFunctions.js`.

psvenk (approved):
Sorry for the repeated comments. After 3f2819c3bb, I think the status quo with the backwards-compatibility guarantee is the way to go, given that it's the least disruptive solution. @jadebuckwalter, feel free to squash-and-merge if you agree.

Issue #187, opened 2020-11-25
Think about changing how GPA and quarter drop downs work


kdk1616:
- Should quarter dropdown be next to year?
- Should GPA be separated from quarter dropdown?
- Should GPA be put into classes table?
- Should we move the year dropdown so that it is clear that this only affects the "Grades" tab? (from #287)

<img width="1434" alt="Pasted Graphic copy" src="https://user-images.githubusercontent.com/31357347/100271495-8ebe5380-2f27-11eb-86e6-c74eb6bfe6bd.png">

psvenk:
On hold until #152 is closed.

psvenk:
The newer mockup that @kdk1616 put in #266: <img width="863" alt="IMG_1458" src="https://user-images.githubusercontent.com/31357347/107558289-95647a80-6ba8-11eb-82fc-44335d204b28.png">

jshaw-7:
for 2.9.1- change dropdown menus to be "year select" and "quarter select", then on the next line (with reset and export buttons) add "Current Q1 GPA: 100" for example, with current/previous and the correct quarter. Move GPA Type button to the next line as well

Issue #116, opened 2020-04-01
Multi-year cumulative GPA


psvenk:
After implementing #74 and #99, add a feature which would use data from multiple years (some of which may be from imported data) to calculate a cumulative GPA.

MohanMaker:
From #287: "The concept of "cumulative GPA" is confusing in the previous year: if it works properly, it shows the average GPA for that year; however, the wording seems to suggest that it shows the cumulative GPA for the past two years, or for all four years. Maybe we can call it "yearlong GPA" or something similar?"

PR #287, opened 2021-04-03
Add support for previous year grades

### LINKED ISSUE #99: Get grades from previous year

psvenk:
Closes #99. Thanks to the REST API, the backend implementation here is much simpler than that in #182. While fetching grades  seems to work, I am marking this pull request as a draft because there are a few known bugs:
- [x] Assignment statistics do not work for previous-year classes
- [x] Switching between current year and previous year seems to cause some buggy behavior in the term dropdown:
image
- [x] The schedule and reports callbacks send data to `currentTableData`, which may be current or previous, meaning that the other option will not have schedule/reports data; the callbacks should be updated to send their data to both `tableData[0]` and `tableData[1]`.
- [x] Switching between years may cause the quarter displayed in the quarter/GPA dropdown to be inconsistent with the quarter whose grades are displayed. This is because, when switching years (except when switching to "previous year"), Aspine tries to look up the same quarter in the other year, which may not exist; in this case, the dropdown displays the nonexistent quarter and the table displays something else.

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -114,6 +114,7 @@
       <select id="tableData_select">
         <option value="0">Current Year</option>
         <option value="0">Current Year</option>
+        <option value="1">Previous Year</option>
         <option value="import">Import Data...</option>
       </select>
     </div>
--- a/public/js/buttonFunctions.js
+++ b/public/js/buttonFunctions.js
@@ -352,7 +352,10 @@ let exportTableData = async function(prefs) {
             const response = await $.ajax({
               url: "/data",
               method: "POST",
-              data: { quarter: parseInt(term.match(/\d+/)[0]) },
+              data: {
+                quarter: parseInt(term.match(/\d+/)[0]),
+                year: "current",
+              },
               dataType: "json json"
             });
             currentTerm = term;
@@ -437,7 +440,7 @@ let importTableData = async function(obj) {
     + `https://github.com/Aspine/aspine/releases</a>.`;
   }
 
-  currentTableDataIndex++;
+  currentTableDataIndex = tableData.length;
   tableData[currentTableDataIndex] = {};
   currentTableData = tableData[currentTableDataIndex];
   currentTableData.type = "imported";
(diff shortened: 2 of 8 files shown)


psvenk:
> - The schedule and reports callbacks send data to `currentTableData`, which may be current or previous, meaning that the other option will not have schedule/reports data; the callbacks should be updated to send their data to both `tableData[0]` and `tableData[1]`.

Fixed in 157bc2d5f5.

psvenk:
> - Assignment statistics do not work for previous-year classes

Fixed in cb14a441ac.

psvenk:
> - Switching between current year and previous year seems to cause some buggy behavior in the term dropdown

Fixed in 55fa4334ca.

psvenk:
> - Switching between years may cause the quarter displayed in the quarter/GPA dropdown to be inconsistent with the quarter whose grades are displayed. This is because, when switching years (except when switching to "previous year"), Aspine tries to look up the same quarter in the other year, which may not exist; in this case, the dropdown displays the nonexistent quarter and the table displays something else.

Fixed in 6e8037ac25.

psvenk:
This should be ready for review. As with #220 (albeit to a lesser extent), these are fairly major changes, so it would be good to have reviews from multiple people before this is merged.

jadebuckwalter (approved):
Looks good to me!

MohanMaker (commented):
Overall looks good to me, thanks for this great feature @psvenk! 

One bug I discovered is that you can not switch the GPA type when the previous year is selected. 

Also, the concept of cumulative GPA is a bit confusing when you are in the previous year since it still shows your cumulative GPA up to the current date. Maybe it could tell you your average GPA for that year, as well as the overall cumulative GPA?

Also, I wish it would show previous year's schedule and attendance reports, but that can come later...

psvenk:
@jadebuckwalter @MohanMaker Thank you for your speedy reviews.

It seems that the GPA code is looking for a property in the previous year data that is not present, causing these two bugs; it should (hopefully) be a simple fix. More generally, this reminds me of @tektaxi's comment that the concept of "cumulative GPA" is more generally confusing in the previous year: if it works properly, it shows the average GPA for that year; however, the wording seems to suggest that it shows the cumulative GPA for the past two years, or for all four years, as is being tracked in #116. Maybe we can call it "yearlong GPA" or something similar?

With respect to the previous year's schedule and attendance reports, I don't think Aspen exposes this information. Maybe we should move the year dropdown so that it is clear that this only affects the "Grades" tab, but this is probably part of #187.

psvenk:
@MohanMaker Both of the bugs that you noticed should now be fixed (the GPA type bug in 65d9a45cb6 and the cumulative GPA bug in 3a168dbbc2); could you check if this is the case on your end when you have some time?

MohanMaker (approved):
Just tested and everything is working correctly! I will also make a comment on the other issues you mentioned above so we can remember to discuss the dropdown and the wording of cumulative GPA.

tektaxi (approved):
Works well for me. I don't see any remaining bugs. We should definitely clarify the cumulative GPA thing though, right now it is still confusing.

Issue #288, opened 2021-04-06
Add support for anticipated (early May) schedule


psvenk:
The school has released the anticipated schedule for the further expansion of in-person learning in accordance with the DESE guidelines, currently slated to go in effect ~some time in late April~ on May 10. We should update Aspine to deal with this schedule properly.

Issue #95, opened 2020-01-14
Figure out what lunch students have


kdk1616:
* [x] Use period three class code to determine what lunch students have
* [x] Show this on schedule

notrodes:
What is the status of this? I thought this was implemented

psvenk:
The current implementation has some flaws. We will need to re-evaluate after some time.

andOrlando:
Not entirely sure if we can be 100% accurate. Don't some classes not conform to the general lunch thing?

psvenk:
The goal is not to be 100% accurate but to go by the policy set by the school, which most classes follow. In any case, this is on hold until we reopen in person.

psvenk:
This is no longer on hold because the new schedule includes two lunches (Lunch A and Lunch B), so we should try to get this in version 2.8.0 along with #288.

psvenk:
This is on hold until the school determines how Lunch A and Lunch B will be assigned.

psvenk:
Lunch assignments were announced today, so this was implemented in 84fc380fcd as part of #294.

PR #290, opened 2021-04-07
Migrate schedule.json to TOML

### LINKED ISSUE #253: Improve formatting of schedule.json
Currently, the start and end times for periods in `public/schedule.json` are stored in milliseconds since midnight. However, we do not need to store information down to seconds, much less milliseconds. It would be more maintainable to store times as number of minutes since midnight (`hour * 60 + minute`). Or, even better, we could switch to [TOML](https://toml.io/en/), which is easier to read and write than JSON and offers native support for dates and times.

psvenk:
Closes #253. This helps readability and maintainability considerably. Some changes have been made to the frontend to work with native JavaScript `Date` objects instead of numbers.


notrodes:
Can you add an error throw as a default case in the switch statement on line 178 of clock.js?

tektaxi (approved):
Works fine for me, but I haven't looked through the code much.

psvenk:
@notrodes Are you referring to the `update_lunch` function? That is currently dormant code which will be addressed in #95 / #288 and is out of scope here. Are there any issues with the code that I wrote/touched in this pull request, or is this ready to merge?

public/js/clock.js
```diff
@@ -29,6 +29,18 @@ let date_override = undefined;
 
 function schedulesCallback(response) {
     schedules = response;
+    // Convert all start and end times to JavaScript Date objects
+    for (const [, schedule] of Object.entries(schedules)) {
+        for (entry of schedule) {
+            // Convert each entry to a JavaScript Date object
+            let [, hours, minutes, seconds, ms]
```
notrodes: Can you avoid this regex repeat?

psvenk: Fixed in 2571fc5157.

notrodes (approved):

PR #292, opened 2021-04-11
Remove VIP username list

psvenk:
This is a largely unused feature which arguably makes Aspine less useful; anyway, all students in the list have already graduated.


psvenk:
@notrodes Why did you close this pull request?

PR #293, opened 2021-04-11
Modularize and strip down PDF viewer

### LINKED ISSUE #175: Use regular PDF.js viewer
Instead of rolling our own UI, consider using the regular PDF.js viewer (demo at <https://mozilla.github.io/pdf.js/web/viewer.html>) which is more full-featured and responsive.

### LINKED ISSUE #289: Break home.js into component tabs
This would increase readability, maintainability, and potentially performance with on the fly module loading.

notrodes:
Also cleanup


notrodes:
Don't be fooled, most of the deletes come from the package.lock file.

psvenk (changes requested):
I know that this is still in a draft state, but this pull request, in its current trajectory, does not seem like it will be in good shape to be merged, for the following reasons:
- Removing pdf.js and relying on the browser causes the Reports tab to be non-functional in environments that do not have built-in PDF viewers. For example, [qutebrowser](https://qutebrowser.org/) does not have a facility to view PDFs inline on a webpage, causing the Reports tab to be completely broken on qutebrowser.
- The controls provided (next page, previous page, zoom, fullscreen, etc.) are completely dependent on the browser. Sometimes, there are no controls at all; for example, iOS Safari shows the first page of a PDF and does not permit any sort of interaction.
- The UI is inconsistent between platforms. Given that this is the Web, we must allow for some variability; however, relying on a component whose implementation and features are highly variable between browsers makes this more difficult to maintain and troubleshoot.
- Using a plain HTML `<select>`, removing the borders, and other design decisions here have a detrimental effect on the UI and UX.

Instead, I propose that you open two separate pull requests: one strictly moving the PDF-related code into its own file (preferably named `public/js/reports.js` or similar to avoid having the same name as the `pdf.js` library), and another addressing #175 and switching to the built-in PDF.js viewer to avoid implementing our own UI while also avoiding the drawbacks of falling back to the browser's built-in facilities (or lack thereof).

PR #294, opened 2021-04-12
Add support for May-June 2021 schedule

### LINKED ISSUE #95: Figure out what lunch students have

### LINKED ISSUE #288: Add support for anticipated (early May) schedule

### LINKED ISSUE #303: Move lunch switch
Right now the switch to change your lunch is on the clock tab, whereas it makes more sense for it to be with the rest of the schedule on the Schedule tab.

psvenk:
Fixes #288. This pull request adds support for the anticipated schedule for May and June 2021. Please note that the current support is preliminary and needs changes in the code.

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -980,9 +980,9 @@ p#large_clock_period {
   top: -44px;
   height: 0px;
   pointer-events: none;
-  letter-spacing: 3.15em;
+  letter-spacing: 7em;
   text-align: center;
-  left: 40px;
+  left: 87.5px;
 }
 
 .unselectable {
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -49,6 +49,7 @@ body {
 
   /* Different variable for gray subtitles that turn to white for dark mode */
   --gray-to-white: #666666;
+  --gray3-to-white: #333;
 
   --black: #000000;
   --transparent-black: rgba(0, 0, 0, 0.6);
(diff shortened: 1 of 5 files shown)


psvenk:
I think this should be ready for review (and hopefully merge). As far as I remember, the only omission here is that lunch guessing has not been incorporated yet. Because we do not yet have information about how lunches will be assigned, we can address that in a separate pull request (to close #95) and merge this for now (which defaults to lunch A).

I'm not sure if this uses `localStorage` to store the user's lunch; someone (possibly me) should test that.

jadebuckwalter:
@psvenk We could temporarily create a lunch slider where people can choose between lunch A and lunch B (like in the past), but that might not be necessary if we implement lunch guessing as soon as we know how lunches are being assigned.

Also, since Wednesdays alternate between black and silver, how should we take that into account? Maybe instead of determining whether it's a black or silver week, we just display both classes (if there are different classes for black and silver days) in Wednesday's schedule and clock. We could also hard-code for the remainder of the year, but that's not the best solution.

jadebuckwalter (approved):
Looks good to me!

PR #297, opened 2021-04-14
Redesign tables on mobile

### LINKED ISSUE #167: Display more information for assignments without statistics

jadebuckwalter:
Create a table for assignments without stats that shows your score, the
maximum score, date assigned, date due, and assignment feedback.
For now, I'm trying to get more data to show up in the modal to remove
table columns on mobile, but eventually, the info in the modal will be
transferred to a separate region under the row in the table.

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -839,7 +839,8 @@ canvas#large_clock {
   width: 250px
 }
 
-#stats_modal_table {
+#stats_modal_table, 
+#no_stats_modal_table {
   font-family: sans-serif;
   border-collapse: collapse;
   width: 100%;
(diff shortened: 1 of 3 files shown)


jadebuckwalter:
Still left to do:
- [ ] Add category dropdown to stats modals on mobile
- [ ] Adjust other tables (grades, categories, schedule) for mobile
- [ ] Make the adjustments dynamic
- [ ] Move info in stats modals to an expanded area in the assignments table

psvenk (commented):
This looks great on the whole. I agree that this is in a mergeable state and that the additional changes should be made down the line.

One point of comment: I see that in the case that there are no assignment statistics (but the basic info — score, max score, comments — exists), there is no message saying "there are no stats". Would it be a good idea to alert the user of the fact that there are no assignment statistics, but this is some expanded information about the assignment?

Additionally, the tooltip on the info icon should be changed from "Statistics" to "Info" to indicate its true purpose.

psvenk (approved):
Looks good to me on the whole once again; thanks for the additional changes. Another issue that I noticed is that the media matching is done only once, at the page load. Is there any way (perhaps via an event listener) to make it dynamic, so that resizing a browser window or rotating a phone would cause the table to adapt?

I suppose that we can address this in a separate pull request, along with the other areas that you identified, so this should be ready to merge.

Edit: Never mind; I just realized that you already identified this issue when you wrote "make the adjustments dynamic".

PR #305, opened 2021-07-02
Small edits to login.js

notrodes:
In this vein, either have server send full error text or server side render could be improvements.

### DIFF
--- a/public/css/login.css
+++ b/public/css/login.css
@@ -120,10 +120,6 @@ textarea {
   border: none;
 }
 
-textarea:focus, input:focus {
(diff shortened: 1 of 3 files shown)


psvenk (approved):
Looks good to me!

psvenk:
Also @notrodes re your comment: my reasoning for having the server send only an error code was that it plays nicely with the enum defined in the server code and that it allows us to change the verbiage without affecting the API in any way. I don't think the client-side performance hit is enough to merit SSR here (though it may be a good idea all things considered).

PR #306, opened 2021-07-26
Edits to snackbar for concision and readability.

notrodes:
3 biggest changes are:

1. using library function Object.assign() instead of multiple assignments from options to this,
2. fixed jsdoc,
3. changed code to use more default params instead of undefined tests.

### DIFF
--- a/public/js/snackbar.js
+++ b/public/js/snackbar.js
@@ -43,7 +43,7 @@ class Snackbar {
         this.text = text;
         Object.assign(this, options);
 
-        switch(options.timeoutMode) {
+        switch (options.timeoutMode) {
             case "hide":
                 this.timeoutEndFunction = () => this.hide();
                 break;
@@ -107,7 +107,7 @@ class Snackbar {
             //creates the text span
             const buttonTextNode = document.createElement("SPAN");
             buttonTextNode.textContent = this.buttonText;
-            // debugger;
+
             //colors the text if necessary
             if (this.textColor)
                 buttonTextNode.style.color = this.textColor;
--- a/public/js/snackbar.js
+++ b/public/js/snackbar.js
@@ -25,27 +25,30 @@ class Snackbar {
     static SHOWN = 2
 
     /**
-     * text is the main requirement, and it's just text
-     * color: String - string reference to a color or a variable, sets the background color
-     * textColor: String - string reference to a color or a variable, sets the text color
-     * buttonText: String - Sets the button text, both it and buttonClick have to be defined for the button to show
-     * buttonClick: Function - Sets the button's onclick logic, both it and buttonText have to be defined for the button to show
-     * destroyWhenButtonClicked : Boolean - Whether or not it should destroy itself when the button is clicked, defaults to true
-     * bodyClick: Function - Sets the body's onclick logic
-     * destroyWhenBodyClicked : Boolean - Whether or not it should destroy itself when the body is clicked, defaults to true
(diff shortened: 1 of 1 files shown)


public/js/snackbar.js
```diff
@@ -39,35 +39,27 @@ class Snackbar {
...
         this.timeoutInProgress;
 
         //what to run on timeout
         this.timeoutEndFunction;
-        switch(options["timeoutMode"]) {
-            case "destroy":
-            case undefined:
-                this.timeoutEndFunction = () => this.destroy();
-                break;
+        switch(options.timeoutMode) {
```
psvenk: We should use a space after `switch` because it is a keyword and not a function (in the same vein as `if`, `for`, `while`, etc.)
```suggestion
        switch (options.timeoutMode) {
```

public/js/snackbar.js
```diff
@@ -249,17 +241,14 @@ class Snackbar {
...
     createID() {
-        let id = null;
-
-        //goes through all consecutive numbers to find an id
-        let iterator = 0
-        while (id === null) {
-            //checks if the id already exists, otherwise continues to iterate
-            Snackbar.snackbarIDs.includes(iterator) ? iterator++ : id = iterator;
-        }
+        debugger;
```
psvenk:
```suggestion
```

psvenk (commented):
This looks good so far (AFAICT; obviously @andOrlando is the authority on snackbar issues). I know that this is still in a draft stage, but I have a couple of comments.

public/js/snackbar.js
```diff
@@ -113,35 +93,32 @@ class Snackbar {
...
-        if (this.buttonText !== undefined && this.buttonClick != undefined) {
+        if (this.buttonText  && this.buttonClick) {
             //creates the button and adds class
             const buttonNode = document.createElement("BUTTON");
 
             //creates the text span
             const buttonTextNode = document.createElement("SPAN");
             buttonTextNode.textContent = this.buttonText;
-
+            // debugger;
```
psvenk:
```suggestion

```

public/js/snackbar.js
```diff
@@ -249,17 +241,14 @@ class Snackbar {
...
-        //goes through all consecutive numbers to find an id
-        let iterator = 0
-        while (id === null) {
-            //checks if the id already exists, otherwise continues to iterate
-            Snackbar.snackbarIDs.includes(iterator) ? iterator++ : id = iterator;
-        }
+        debugger;
+        let id = 0;
+        // goes through all consecutive numbers to find an id
+        for (; id in Snackbar.snackbarIDs; id++); // checks if the id already exists, otherwise continues to iterate
```
psvenk, line 224: Could we just do something like this?
```suggestion
        const id = Math.max(...Snackbar.snackbarIDs) + 1;
```

andOrlando:
> This looks good so far (AFAICT; obviously @andOrlando is the authority on snackbar issues). I know that this is still in a draft stage, but I have a couple of comments.

It looks good to me with your comments

public/js/snackbar.js
notrodes: Theoretically the IDs could become fragmented as objects are destroyed.

psvenk: Good point; you're right that the `for` loop prevents fragmentation.

psvenk (approved):
Looks good to me.

PR #311, opened 2021-09-28
Add support for fall 2021 schedule

### LINKED ISSUE #308: Add support for 2021-22 schedule
Add support for the schedule currently in use for fall 2021.

Cc @tektaxi

psvenk:
Closes #308 and adds support for the new schedule to `schedule.toml` and the frontend generally. Lunch detection has not yet been updated for the new system.

Additionally, there is a bug on Aspen which causes periods to be misaligned. We could compensate for this with some hacky code in our backend, but this will break once Aspen is fixed. So, I think the best course of action is to keep the backend code as-is and wait for Aspen to be fixed.
image

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -940,7 +940,7 @@ p#large_clock_period {
 }
 #lunch_range {
   -webkit-appearance: none;
-  width: 100px;
+  width: 157.5px;
   height: 34px;
   border-radius: 17px;
   background: var(--gray3-to-white);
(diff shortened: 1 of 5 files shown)


psvenk:
The lunch issue has (hopefully) been fixed, so this is ready for review (@tektaxi and/or anyone else).

tektaxi (approved):
works great for me! this is great for getting Aspine ready for this year.

Issue #102, opened 2020-02-21
Improving Grade Prediction Algorithm


Ruborcalor:
Currently grade predictions are calculated exclusively using the "Category Total Points" strategy, however teachers have the option to choose between four different grade calculation styles: "Categories", "Categories and Assignments", "Total Points", and "Category Total Points". Find explanations for each of these in the attached [AspenAverageModes.pdf](https://github.com/Aspine/aspine/files/4236281/AspenAverageModes.pdf)

It would be great to find a way to induce which of the above styles a teacher has selected for a class, most likely by trying each method and seeing which one comes closest to a student's actual grade in a class.

notrodes:
I consider this a bug since it makes the core functionality of Aspine worse.

psvenk:
From #313 (closed as duplicate):
<blockquote>

Some teachers do not use categories on Aspen and just have one category for all assignments with weight "N/A". Aspine currently chokes on this and turns it into NaN, causing bad results when the user tries to edit assignments. Instead, we should treat an N/A-weighted category as weight 100%.

<details>
<summary>
Category list in Aspen
</summary>

![image](https://user-images.githubusercontent.com/45520974/137819225-08898063-3cf8-43fa-a650-5268cf4c598f.png)

</details>
<details>
<summary>
Category list in Aspine
</summary>

![image](https://user-images.githubusercontent.com/45520974/137819280-ec1f7b9e-53c1-485c-a196-2ea18e471b44.png)

</details>
<details>
<summary>
Class average after making an edit
</summary>

![image](https://user-images.githubusercontent.com/45520974/137819335-57affc9a-255e-48b9-bd91-6c58488c5e48.png)

</details>
</blockquote>

psvenk:
![image](https://user-images.githubusercontent.com/45520974/152695697-020ac7bb-ef93-48e7-b230-dfe54d8a8734.png)

This is from the `pastDue` endpoint of the Aspen API, suggesting that there is specific support for standards-based grading in Aspen. We should look into supporting this in Aspine.

PR #314, opened 2021-10-19
Handle "N/A" category weight

### LINKED ISSUE #313: Properly handle N/A category weights
Some teachers do not use categories on Aspen and just have one category for all assignments with weight "N/A". Aspine currently chokes on this and turns it into NaN, causing bad results when the user tries to edit assignments. Instead, we should treat an N/A-weighted category as weight 100%.

<details>
<summary>
Category list in Aspen
</summary>

image

</details>
<details>
<summary>
Category list in Aspine
</summary>

image

</details>
<details>
<summary>
Class average after making an edit
</summary>

image

</details>

psvenk:
Closes #313. A category weight of "N/A" is now properly handled as a category with weight 100% (presuming that there are no other categories).

### DIFF
--- a/src/scrape.ts
+++ b/src/scrape.ts
@@ -73,7 +73,7 @@ export async function get_student(
       }
 
       let categories: { [key: string]: string } = {};
-      for (const [cat, { weight} ] of details.categories) {
+      for (const [cat, { weight } ] of details.categories) {
         categories[cat] = weight.toString();
       }
       return {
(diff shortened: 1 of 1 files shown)


jadebuckwalter (commented):
I'm not sure how to properly test this, because I don't have any classes this year or last year for which this is the case. However, I did have a class last year with multiple unweighted categories, and those still show up as "NaN%". Of course, the changes you made only apply if there is only one category, but this might be something to consider fixing as well.

psvenk:
@jadebuckwalter You bring up a good point. In fact, I think properly solving #313 requires solving #102 to detect the "Total Points" strategy and handle it properly.

PR #315, opened 2021-10-20
Update README to reflect info tab verbiage

psvenk:
Re-word the first two sections to be more informative, and reflect the language used on the "Info" tab within Aspine.

We should also change the "about" text of the GitHub repository in this vein.


jadebuckwalter (approved):
Looks good to me!

PR #319, opened 2021-10-21
http/2 drop in replacement package for https with spdy for http/2 support.

### LINKED ISSUE #201: Http2
Its faster better cheaper, http2.

notrodes:
Closes #201.

I think it would be good to monitor performance somehow to see if it does not harm it. Otherwise, super simple.


psvenk (changes requested):
I found the cause of my issues running this code: [the `spdy` module is incompatible with recent versions of Node.js](https://github.com/spdy-http2/node-spdy/issues/380). I suggest that we wait until that issue is fixed to add HTTP/2 support.

PR #330, opened 2022-01-31
Redesign tables on mobile

### LINKED ISSUE #203: Categories and Assignments really hard to read on mobile
On mobile, it's impossible to see the full name of an assignment or category with the current state of the table. We should reorganize it on small screens to be more readable. Maybe only display the name and category in the table, and click on the assignment for the score and more details? I'm not sure the specifics, but right now it's kind of a mess.

IMG_65C06BC444EC-1

jadebuckwalter:
Hide and show columns in the Assignments and Categories tables based on
the screen size. Fixes #203.

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -734,8 +734,6 @@ canvas#large_clock {
 
 .stats-window {
   display: none;
-  padding-top: 50px;
-  padding-bottom: 50px;
   width: 100%;
 }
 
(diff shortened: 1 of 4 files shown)


psvenk:
I just now made a commit that moves the stats modal to be between the categories and assignments tables, but the styling could use some work. We should also look into giving the other modals the same treatment.

psvenk (changes requested):
The assignments table header still does not load properly on mobile for me. Apart from that, this looks good to me.

psvenk (approved):
I can no longer reproduce the issue with the assignments table header; maybe there was some caching issue going on with my previous test. So, I think we can consider that issue fixed and this pull request ready to merge.

PR #331, opened 2022-02-01
Forwards and backwards

### LINKED ISSUE #224: Router, using the back and forwards keys
to do spa's right we should integrate a router.

notrodes:
Closes #224 and it was way easier than I thought it would be.

### DIFF
--- a/public/home.html
+++ b/public/home.html
@@ -88,14 +88,14 @@
 
   <div class="tab">
     <button class="tablinks hide switch-exempt" onclick="openSideNav()" id="hamburger_button"><i class="fa fa-bars"></i></button>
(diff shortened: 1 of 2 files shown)


psvenk (approved):
Looks good to me. One thing to consider: we could assign URL fragments to the history states — e.g., `#schedule`, `#attendance` — and switch to the proper tab if a user loads the page with one of these fragments as part of the URL. This makes the tabs function more like separate pages and has the benefit of making the history states transparent to the user (via changes in the page URL). What are your thoughts?

notrodes:
The thing with that is that I couldn't come up with a good way to translate between #fragment names and ids for lookup. If you think of a way of gathering and selecting elements by #'s that would be great but I can just brute force it, in a not very programmatic translator, with a series of ifs or a switch.

PR #333, opened 2022-02-06
Re-implement recent activity

### LINKED ISSUE #180: Re-enable recent activity page for assignments
Aspen no longer displays recently inputted grades, but it could still be useful to have this feature. Aspine could look for assignments that have a recent "dateasgn" or "datedue", and display those. It could also display when term grades are posted.

notrodes:
Closes #180. Its 60% working, things that don't work:

- Colors
- max scores and percentage
- clicking on rows


notrodes:
Its 60% working, things that dont work:

- Colors
- max scores and percentage
- clicking on rows

src/scrape.ts
```diff
@@ -360,7 +360,7 @@ async function get_recent(session: Session): Promise<Recent> {
   const recentActivityArray =
     [...document.querySelectorAll("gradebookScore")].map(x =>
       Object.fromEntries([
-        "date", "classname", "score", "assignment",
+        "date", "classname", "grade", "assignmentname",
```
psvenk: This change, along with the associated change on the frontend, causes unneeded API breakage. The names might not be ideal, but it's better just to leave them as is, so that we can avoid incompatibilities.

public/js/home.js
```diff
@@ -185,8 +186,8 @@ let recentActivity = new Tabulator("#recentActivity", {
     columns: [
         {title: "Date", field: "date", formatter: rowFormatter},
         {title: "Class", field: "classname", formatter: classFormatter},
-        {title: "Assignment", field: "assignment", formatter: rowFormatter, headerSort: false},
-        {title: "Score", field: "score", formatter: rowFormatter, headerSort: false},
+        {title: "Assignment", field: "assignmentname", formatter: rowFormatter, headerSort: false},
+        {title: "Score", field: "grade", formatter: rowFormatter, headerSort: false},
         {title: "Max Score", field: "max_score", formatter: rowFormatter, headerSort: false},
         {title: "Percentage", field: "percentage", formatter: rowGradeFormatter},
     ],
```
psvenk: See comment on scraper.

public/js/home.js
```diff
@@ -274,8 +275,8 @@ let mostRecentTable = new Tabulator("#mostRecentTable", {
         //{title:"Date", field:"date", formatter: rowFormatter, headerSort: false},
         {title: "Date", field: "date", formatter: rowFormatter},
         {title: "Class", field: "classname", formatter: classFormatter},
-        {title: "Assignment", field: "assignment", formatter: rowFormatter, headerSort: false},
-        {title: "Score", field: "score", formatter: rowFormatter, headerSort: false},
+        {title: "Assignment", field: "assignmentname", formatter: rowFormatter, headerSort: false},
+        {title: "Score", field: "grade", formatter: rowFormatter, headerSort: false},
         {title: "Max Score", field: "max_score", formatter: rowFormatter, headerSort: false},
         {title: "Percentage", field: "percentage", formatter: rowGradeFormatter},
     ],
```
psvenk: See comment on scraper.

psvenk (changes requested):

notrodes:
They are necessary changes because Aspen changed the names of the where the data is stored.

psvenk (dismissed):
Looks mostly good to me. The only thing left to do here is to make sure that the max score and percent for recent assignments are filled out properly. We could do this by looking up the assignment in the grades table using the class name and assignment name.

jshaw-7:
not sure if this is purposeful, but the recent activity only shows up when you first load the grades page. if you click on a specific class and then hit the reset button, the recent activity is gone. also not sure how i feel about the color scheme, colors based on the grades might be better than doing it by class.

PR #343, opened 2022-10-02
fixed predicted grades being inaccurate

brianliu8709:


tektaxi (approved):
looks good! Excited for my grades to be more accurate

PR #345, opened 2022-10-13
login links

jshaw-7:
Fixed old links and their formatting on the Login page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "aspine",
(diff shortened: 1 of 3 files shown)


tektaxi (changes requested):
I like this! I think we could reduce the space between the district logo and the line with all the links, it feels like there's a lot of emptiness. Also not sure about the Instagram link...

PoobyMalom (approved):
Approved

jshaw-7:
>

tektaxi (approved):
looks good! great work @jshaw-7 🎉

PR #347, opened 2022-10-19
Change reset/export button locations. Helps with mobile (viewing calculated grades)

jshaw-7:
changed the main grades tabulator and moved reset/export buttons to better see calculated grades on smaller screens


tektaxi (approved):
this is gonna make Aspine look so good!

PR #360, opened 2023-05-13
Recent activity

### LINKED ISSUE #180: Re-enable recent activity page for assignments

jshaw-7:
added recent activity table on grades tab. changed attendance tab to 'attendance & recent' and displays recent activity with a switch.


tektaxi (approved):
Looks good! This is a much-needed addition. 
We can try to add more fields to the recent activity section by matching the assignments by ID in a later update if we would like.

PR #361, opened 2023-05-13
added text to export data and reset buttons

### LINKED ISSUE #318: Add text to clarify use of buttons
Particularly for the "export" and "reset" buttons, add text so that it is more clear what these buttons do. (This would only apply to desktop, not to mobile.)

jshaw-7:


tektaxi (approved):
Personally, I would opt for a less bold font next to the buttons, I think it's a little strange as is. Other than that, looks great!

PR #366, opened 2024-05-22
Added script to run the webapp as a service

SwaroopGiri:
added enable-service.sh which can be run once as a post installation script to make the web application persistant across updates and reboots


ABLavine (approved):
LGTM

PR #368, opened 2024-05-22
Easter eggs

ThomasM-R:
super important vital changes that cannot be forgotten
also @splice-7 is me, i just switched accounts, so someone might want to give me write access to the repo
i reacted with 👍 with that account just in case ¯\\\_(ツ)\_/¯

### DIFF
--- a/public/css/home.css
+++ b/public/css/home.css
@@ -26,6 +26,15 @@ body, html {
   font-family: Poppins-Regular, sans-serif;
 }
 
+body.dank {
+  animation: dankmode 1s linear infinite;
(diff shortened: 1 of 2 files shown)


bufftofubrudder (approved):

splice-7 (commented):
this requires testing in aspine v3.0.0
in theory this should have no conflicts with any new features in v3 as it is purely aesthetic, but just to be sure this should be tested
to test: open the aspine main page, type "dankmode" and enter the konami code (↑↑↓↓←→←→B A)

